PR #5, opened 2021-10-28
feat: add git-lfs support

benoitf:
### What it does
Currently if you clone https://github.com/Apress/repo-with-large-file-storage it won't clone LFS files you'll see files with few bytes

And if you try with the updated image, files are there as expected

### how to test
example of repository to clone: https://github.com/Apress/repo-with-large-file-storage

### Related issue
https://github.com/eclipse/che/issues/20642

### DIFF
--- a/base/ubi8/Dockerfile
+++ b/base/ubi8/Dockerfile
@@ -20,7 +20,7 @@ LABEL io.openshift.expose-services=""
 USER 0
 
 RUN dnf update -y && \
-    dnf install -y bash curl diffutils git iproute jq less lsof man nano procps \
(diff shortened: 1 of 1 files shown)


svor (approved):
Project was cloned inside the image without problem:
```
bash-4.4$ git clone https://github.com/Apress/repo-with-large-file-storage
Cloning into 'repo-with-large-file-storage'...
remote: Enumerating objects: 16, done.
remote: Total 16 (delta 0), reused 0 (delta 0), pack-reused 16
Unpacking objects: 100% (16/16), 146.83 KiB | 839.00 KiB/s, done.
```

PR #6, opened 2021-11-01
chore(build): fix docker build on linux

vitaliy-guliy:
Signed-off-by: Vitaliy Gulyy <vgulyy@redhat.com>

Fixes build of universal image on linux

### How to test
- try to build the universal image on linux systems
```
$ cd universal/ubi8
$ DOCKER_BUILDKIT=1 docker image build --progress=plain -t quay.io/devfile/universal-developer-image:ubi8-latest .
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,6 @@ Containers images with tools for developers 👨‍💻👩‍💻
 
 ### Red Hat Universal Base Image ([UBI](https://developers.redhat.com/articles/ubi-faq#)) based image ([quay.io/devfile/base-developer-image:ubi8-latest](https://quay.io/repository/devfile/base-developer-image/))
 
-Build with Docker buildkit:
-
-```bash
-$ cd base/ubi8
-$ DOCKER_BUILDKIT=1 docker image build --progress=plain -t quay.io/devfile/base-developer-image:ubi8-latest .
-```
-
 Run the following command to test it with Docker:
 
 ```bash
@@ -68,13 +61,6 @@ $ docker run -ti --rm \
 ## Developer Universal Image
 
 ### Red Hat Universal Base Image ([UBI](https://developers.redhat.com/articles/ubi-faq#)) based image ([quay.io/devfile/universal-developer-image:ubi8-latest](https://quay.io/repository/devfile/universal-developer-image/))
-
-Build with Docker buildkit:
-
-```bash
-$ cd universal/ubi8
-$ DOCKER_BUILDKIT=1 docker image build --progress=plain -t quay.io/devfile/universal-developer-image:ubi8-latest .
-```
 
 Run the following command to test it with Docker: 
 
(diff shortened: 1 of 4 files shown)


universal/ubi8/Dockerfile
```diff
@@ -16,7 +16,7 @@ LABEL description="Image with developers tools. Languages SDK and runtimes inclu
 LABEL io.k8s.display-name="devfile-developer-universal"
 LABEL io.openshift.expose-services=""
 
-USER 1001
+USER 10001
```
svor, line 19: do we really need these changes?

svor (approved):
LGTM

universal/ubi8/Dockerfile
vitaliy-guliy: I suppose `1001` here is a typo

vitaliy-guliy: I wonder if the real user ID in a container is 1234, why do we use 10001 instead?
```
bash-4.4$ id     
uid=1234(user) gid=0(root) groups=0(root)
```

l0rd:
We are mixing how to use it (for final users) with how to build it (for contributors). I think we should move the build instructions in a CONTRIBUTING.md file (like https://github.com/kubernetes/kubernetes-template-project/blob/main/CONTRIBUTING.md).

base/ubi8/entrypoint.sh
```diff
@@ -1,4 +1,12 @@
 #!/bin/bash
+#
+# Copyright (c) 2021 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
```
l0rd, line 9: We are not using EPL for this project but [Apache 2.0](https://github.com/devfile/developer-images/blob/main/LICENSE)

README.md
```diff
@@ -8,6 +8,13 @@ Containers images with tools for developers 👨‍💻👩‍💻
 
 ### Red Hat Universal Base Image ([UBI](https://developers.redhat.com/articles/ubi-faq#)) based image ([quay.io/devfile/base-developer-image:ubi8-latest](https://quay.io/repository/devfile/base-developer-image/))
 
+Build with Docker buildkit:
+
+```bash
+$ cd base/ubi8
+$ DOCKER_BUILDKIT=1 docker image build --progress=plain -t quay.io/devfile/base-developer-image:ubi8-latest .
+```
+
```
l0rd, line 17: We are mixing how to use it (for final users) with how to build it (for contributors). I think we should move the build instructions in a CONTRIBUTING.md file (like https://github.com/kubernetes/kubernetes-template-project/blob/main/CONTRIBUTING.md).

universal/ubi8/Dockerfile
l0rd: The user ID is set by the underlying container runtime. It can be a random one (`1234`), the default one (`10001`) or a specific one (`0`). `10001` is a user that is not privileged.

And yes, that's probably a typo.

base/ubi8/entrypoint.sh
vitaliy-guliy: Fixed

README.md
vitaliy-guliy: Done

PR #8, opened 2021-11-12
Add launch scripts for Scala tools

sideeffffect:


universal/ubi8/Dockerfile
```diff
@@ -44,6 +44,17 @@ ENV PATH="/home/user/.sdkman/candidates/maven/current/bin:$PATH"
...
+RUN curl -fLo cs https://git.io/coursier-cli && \
+    chmod +x cs && \
+    mv cs /usr/local/bin/
+RUN curl -fLo sbt https://raw.githubusercontent.com/dwijnand/sbt-extras/master/sbt && \
+    chmod +x sbt && \
+    mv sbt /usr/local/bin/
+RUN curl -fLo mill https://raw.githubusercontent.com/lefou/millw/main/millw && \
+    chmod +x mill && \
+    mv mill /usr/local/bin/
+
```
svor, line 57: This block with installing scala tools should be moved under `USER 0` to avoid docker build failure like `cannot move 'cs' to '/usr/local/bin/cs': Permission denied`

sideeffffect: Oh, good point, thanks. I've moved it :+1:

svor (approved):

PR #9, opened 2021-11-17
chore: install clang toolset and gdb

svor:
Signed-off-by: svor <vsvydenk@redhat.com>

- Install `gdb`, `clangd` and `clang` with libraries
- Remove `git` from universal image since it was installed in base image


benoitf (approved):
👍

PR #11, opened 2021-11-25
chore: Install PHP 7.4 with libraries, composer and xdebug

svor:
Signed-off-by: svor <vsvydenk@redhat.com>

Installs PHP 7.4 with libraries, php composer and xdebug from PHP Extension Community Library

### DIFF
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -81,6 +81,30 @@ RUN dnf -y module enable python39:3.9 && \
     dnf -y install python39 python39-devel python39-setuptools python39-pip nss_wrapper && \
     dnf -y clean all --enablerepo='*'
 
+# PHP
+ENV PHP_VERSION=7.4
+RUN dnf -y module enable php:$PHP_VERSION && \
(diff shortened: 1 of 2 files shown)


universal/ubi8/Dockerfile
```diff
@@ -81,6 +81,30 @@ RUN dnf -y module enable python39:3.9 && \
     dnf -y install python39 python39-devel python39-setuptools python39-pip nss_wrapper && \
     dnf -y clean all --enablerepo='*'
 
+# PHP
+ENV PHP_VERSION=7.4
```
benoitf, line 85: maybe we should use 8.0 as 7.4 is no longer active in a couple of days
https://www.php.net/supported-versions.php

benoitf: it seems anyway 8.0 is not in ubi8 so...

benoitf (approved):

universal/ubi8/Dockerfile
svor: @benoitf why I've decided to install 7.4, because it is the latest version which is used in RedHat images https://catalog.redhat.com/software/containers/search?q=php&p=1

PR #15, opened 2021-12-09
feat: Add Java8 and Java17

benoitf:
default is still Java11

Fixes https://github.com/eclipse/che/issues/20793


svor (approved):

svor:
related issue https://github.com/eclipse/che/issues/20793

PR #24, opened 2022-03-15
feat: use nvm to install node and also provides different nodejs versions

benoitf:
It provides NodeJS 12, NodeJS 14 and 16

16/LTS is the default


l0rd (approved):
👍

nickboldt:
Interesting choice, as Node 12 is EOL in ~ 6weeks ( https://nodejs.org/en/about/releases/ )

This is also not a viable solution for downstream, as we install ONLY from rpms / modules (no curl allowed in Brew).

Ref: https://github.com/eclipse/che/issues/20858

PR #25, opened 2022-05-05
feat: add e2fsprogs library for support IntelliJ IDEA 2021

vzhukovs:
Since `e2fsprogs-static` package has removed from RHEL 8 distribution, it is not possible to install from the repository.

See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/considerations_in_adopting_rhel_8/index#removed-packages_changes-to-packages

There is a proposal to include latest released version of `e2fsprogs`.
`e2p.so` is used by the IntelliJ IDEA [FileSystemUtil.java#L763-L769](https://github.com/JetBrains/intellij-community/blob/master/platform/util/src/com/intellij/openapi/util/io/FileSystemUtil.java#L763-L769).

Part of https://github.com/eclipse/che/issues/21007

Signed-off-by: Vladyslav Zhukovskyi <vzhukovs@redhat.com>

### DIFF
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -378,6 +378,32 @@ cd -
 rm -rf "${TEMP_DIR}"
 EOF
 
+# e2fsprogs setup
+# Since e2fsprogs-static package has removed RHEL 8 distribution, it is not possible to install from the repository
+# https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/considerations_in_adopting_rhel_8/index#removed-packages_changes-to-packages
+RUN <<EOF
+set -euf -o pipefail
+TEMP_DIR="$(mktemp -d)"
+cd "${TEMP_DIR}"
+E2FSPROGS_VERSION="1.46.5"
+E2FSPROGS_TGZ="e2fsprogs-${E2FSPROGS_VERSION}.tar.gz"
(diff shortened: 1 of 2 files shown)


universal/ubi8/Dockerfile
```diff
@@ -378,6 +378,32 @@ cd -
...
 
+# e2fsprogs setup
+# Since e2fsprogs-static package has removed RHEL 8 distribution, it is not possible to install from the repository
+# https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/considerations_in_adopting_rhel_8/index#removed-packages_changes-to-packages
+RUN <<EOF
+set -euf -o pipefail
+TEMP_DIR="$(mktemp -d)"
+cd "${TEMP_DIR}"
+E2FSPROGS_VERSION="1.46.5"
+E2FSPROGS_TGZ="e2fsprogs-${E2FSPROGS_VERSION}.tar.gz"
```
nickboldt, line 389: if you need this in downstream UDI image, you'll need a 'prefetch and reuse' approach. See https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/get-sources.sh#L76-L92

vzhukovs: yep, in downstream UDI this tarball will be prefetched for offline build

benoitf (approved):
maybe README file needs to be updated to reflect this new lib

vzhukovs:
> maybe README file needs to be updated to reflect this new lib

Added a notice about installed library to dedicate section in README.md.

PR #35, opened 2022-09-29
Feat: Small change to add shasum for chectl

SDawley:
Issue: che-incubator/chectl#2239
The chectl package-binaries command uses shasum, so we either have to add shasum to the image used to create a workspace or add a symlink from sha256sum (which we do have).

Steps to reproduce: 
- Create a Chectl workspace in the che dogfooding instance from https://github.com/SDawley/chectl/tree/contribute (my branch that contains the build fixes).
- Build Chectl
- Run the package-binaries command
- command fails with `/bin/sh: shasum: command not found`

Steps to verify:
- Create a Chectl workspace in the che dogfooding instance from https://github.com/SDawley/chectl/tree/contribute. It uses the latest ubi image by default.
- Build Chectl
- Run the package-binaries command

### DIFF
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -152,7 +152,6 @@ ENV KAMEL_VERSION 1.9.2
 RUN curl -L https://github.com/apache/camel-k/releases/download/v${KAMEL_VERSION}/camel-k-client-${KAMEL_VERSION}-linux-64bit.tar.gz | tar -C /usr/local/bin -xz \
     && chmod +x /usr/local/bin/kamel
 
-
 # Cloud
 
 # oc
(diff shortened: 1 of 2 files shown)


universal/ubi8/Dockerfile
```diff
@@ -167,6 +166,9 @@ RUN dnf -y module enable container-tools:rhel8 && \
     dnf -y install podman buildah skopeo fuse-overlayfs && \
     dnf -y clean all --enablerepo='*'
 
+# shasum
+RUN dnf -y install shasum
```
l0rd, line 170: This part of the Dockerfile is the very last and that's where we install all the cloud specific tools. Since that's just a small CLI utility I think the best place would be [at the beginning of the base image](https://github.com/devfile/developer-images/blob/main/base/ubi8/Dockerfile#L23).

SDawley: OK, done.

l0rd (approved):

universal/ubi8/Dockerfile
l0rd: Thank you. The PR is approved. Let me know if you are able to merge it or if you want me to do it.

ibuziuk (approved):
@SDawley it would be really helpful if you could provide a proper description to the PR with steps to verify ;-) 
My understanding is that the only problematic command that is not working for chectl with the latest tag is `yarn pack-binaries` (`/bin/sh: shasum: command not found
`) and all other commands are working correctly.

PR #42, opened 2022-10-26
chore: CRW-3445 backport update to...

nickboldt:
chore: CRW-3445 backport update to downstream UDI image into Che UDI image - remove user and add git branch to prompt

Signed-off-by: Nick Boldt <nboldt@redhat.com>


nickboldt:
See also https://github.com/devfile/developer-images/pull/43, which might be better solution in case entrypoint.sh is overridden?

PR #61, opened 2023-02-15
Add a ubi9 version of Developer Base Image

etsauer:
A few notes:

* for now I've removed the epel repository, and the couple of packages we were pulling from there. They don't appear to be critical path, but we'll see.
*  the two packages missing are p7zip and p7zip-plugins

### DIFF
--- a/.github/workflows/ubi9-build.yaml
+++ b/.github/workflows/ubi9-build.yaml
@@ -1,0 +1,80 @@
+
+name: Build of UBI 9 based Developer Images
+
+on:
+  push:
+    branches: [ main ]
+
+jobs:
+  build_base_ubi9_image:
+    name: Build and publish base ubi9 image to Quay.io
+    runs-on: ubuntu-22.04
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/ubi9-build.yaml
```diff
@@ -0,0 +1,80 @@
...
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+      - name: Docker Build & Push Universal
+        uses: docker/build-push-action@v2
+        with:
+          push: ${{ github.event_name != 'pull_request' }}
+          context: universal/ubi9
+          tags: |
+            quay.io/devfile/universal-developer-image:latest
+            ${{ steps.meta.outputs.tags }}
```
l0rd, line 80:
```suggestion
```

I am afraid that the GH action will fail if you try to build the ubi9 based universal image

l0rd (approved):

PR #66, opened 2023-03-02
Update some package versions and temporarily remove git-lfs to addres…

etsauer:
…s some vulnerabilities in the image

This is mostly just updates to the versions of some packages we're installing.

However, one outstanding issue is git-lfs. It appears that the version of this package currently in the ubi repos is:

```
$ git-lfs --version
git-lfs/2.13.3 (GitHub; linux amd64; go 1.17.7)
```

However, upstream they are at version 3.3.0, which uses an updated go version. go 1.17.7 has a number of known vulnerabilities that should be addressed. I think the next step is probably opening a BZ against rhel9 to see if they can update the package, or what the process looks like to pull in updates.

### DIFF
--- a/base/ubi9/Dockerfile
+++ b/base/ubi9/Dockerfile
@@ -78,7 +78,6 @@ RUN \
     echo "export PS1='\W \`git branch --show-current 2>/dev/null | sed -r -e \"s@^(.+)@\(\1\) @\"\`$ '" >> /home/user/.bashrc && \
     # Copy the global git configuration to user config as global /etc/gitconfig
     #  file may be overwritten by a mounted file at runtime
-    # cp /etc/gitconfig /home/user/.gitconfig && \
     # Set permissions on /etc/passwd and /home to allow arbitrary users to write
     chgrp -R 0 /home && \
     chmod -R g=u /etc/passwd /etc/group /home && \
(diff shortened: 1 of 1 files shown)


l0rd (approved):

base/ubi9/Dockerfile
```diff
@@ -77,7 +78,7 @@ RUN \
     echo "export PS1='\W \`git branch --show-current 2>/dev/null | sed -r -e \"s@^(.+)@\(\1\) @\"\`$ '" >> /home/user/.bashrc && \
     # Copy the global git configuration to user config as global /etc/gitconfig
     #  file may be overwritten by a mounted file at runtime
-    cp /etc/gitconfig /home/user/.gitconfig && \
+    # cp /etc/gitconfig /home/user/.gitconfig && \
```
l0rd, line 81: If not needed we should remove this line rather than commenting this out.

etsauer: Removed.

base/ubi9/Dockerfile
```diff
@@ -77,7 +78,6 @@ RUN \
     echo "export PS1='\W \`git branch --show-current 2>/dev/null | sed -r -e \"s@^(.+)@\(\1\) @\"\`$ '" >> /home/user/.bashrc && \
     # Copy the global git configuration to user config as global /etc/gitconfig
     #  file may be overwritten by a mounted file at runtime
```
l0rd, line 80:
```suggestion
```

l0rd (approved):

PR #74, opened 2023-04-05
chore: update dotnet version to 6.0

svor:
Update dotnet version to 6.0

Related issues: https://github.com/eclipse/che/issues/22077, https://github.com/eclipse/che/issues/21978

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ docker run -ti --rm \
 | `go`                |`go-toolset`                         |
 | `gopls`             |`golang.org/x/tools/gopls`           |
 |--------.NET---------|-------------------------------------|
-| `dotnet`            |`dotnet-sdk-5.0`                     |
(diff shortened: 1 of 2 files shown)


nickboldt:
Why not update to dotnet 7?

https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/Dockerfile#L251-L261

nickboldt (approved):
Why not update to dotnet 7?

https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/Dockerfile#L251-L261

svor:
[free-omnisharp-vscode](https://github.com/muhammadsammy/free-omnisharp-vscode) may work incorrectly with .NET 7. Let's start with .NET6 which is LTS

nickboldt:
is that a different extension from what we use downstream?

PR #107, opened 2023-07-24
feat: add support to run kubedock at workspace startup

l0rd:
This PR partially addresses https://github.com/eclipse/che/issues/20227. 

## Using kubedock to `run` containers and local podman to `build` containers

The approach of this PR is to add [`kubedock`](https://github.com/joyrex2001/kubedock/) in the universal developer image (Eclipse Che default image) and start it automatically if the env variable `KUBEDOCK_ENABLED` is set to `true` (can be set using a devfile).

When `KUBEDOCK_ENABLED=true` then the following commands will be executed with kubedock (the remaining commands, in particular `podman build`, will be executed by the local podman):

```
podman run
podman ps
podman exec
podman cp
podman logs
podman inspect
podman kill
podman rm
podman wait
podman stop
podman start
```

## The limits of this approach

The `kubedock` approach allows to run containers from a container running on Kubernetes but has a few limits:
- `podman build -t <image> . && podman run <image>` doesn't work and should be replaced with `podman build -t <image> . && podman push <image> && podman run <image>`. That's because kubedock runs the container as a Kubernetes pod and, as a result, the kubelet on the node will try to pull `<image>` and will fail if it cannot find it.
- Clients that use the podman or docker API need to be configured to point to kubedock setting `CONTAINER_HOST=tcp://127.0.0.1:2475` or `DOCKER_HOST=tcp://127.0.0.1:2475` when they run containers and configured to point to local `podman` when building the container (something impracticable in some cases).
- Some `podman` comands such as `generate kube` are not supported yet
- If the option `--env` is provided then the command `podman run` fails  

## How to test

I have pushed this image to `quay.io/mloriedo/universal-developer-image:kubedock-wrapper`. 

The PR can be tested on sandbox using the repo `https://github.com/l0rd/dockerfile-hello-world`:

- [with kubedock support disabled (to check that podman build works as before)](https://workspaces.openshift.com/#https://github.com/l0rd/dockerfile-hello-world/?image=quay.io/mloriedo/universal-developer-image:kubedock-wrapper)
- [with kubedock support enabled (to check that podman build/push/run works)](https://workspaces.openshift.com/#https://github.com/l0rd/dockerfile-hello-world/tree/kubedock)

### DIFF
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -216,6 +216,10 @@ curl -sSL -o ~/.kubectl_aliases https://raw.githubusercontent.com/ahmetb/kubectl
 echo '[ -f ~/.kubectl_aliases ] && source ~/.kubectl_aliases' >> /home/user/.bashrc
 EOF
 
+# kubectl completion
+RUN kubectl completion bash > /usr/share/bash-completion/completions/kubectl \
+    && echo "source /usr/share/bash-completion/completions/kubectl" >> /home/user/.bashrc
+
 ## shellcheck
 RUN <<EOF
 dnf install -y xz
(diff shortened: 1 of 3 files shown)


svor (approved):
LGTM

screenshot-devspaces apps sandbox-m4 g2pi p1 openshiftapps com-2023 07 25-17_54_28

deerskindoll:
https://github.com/eclipse-che/che-docs/pull/2710

PR #109, opened 2023-07-25
Revork GitHub workflow without using secrets

musienko-maxim:
* According to the issue: https://github.com/eclipse/che/issues/22364 seems there is one way to avoid the problem:
 - Build image directly in the minikube using connection docker daemon 
 - Change pull policy for DWO and run a test

### DIFF
--- a/.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
+++ b/.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
@@ -19,7 +19,6 @@ on:
       - LICENSE
       - '.rebase/*'
 
-
 env:
    USERSTORY: CloneGitRepoAPI
    TS_API_TEST_KUBERNETES_COMMAND_LINE_TOOL: kubectl
--- a/.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
+++ b/.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
@@ -53,12 +53,12 @@ jobs:
     - name: Build base image
       run: |
        eval $(minikube docker-env)
-       cd base/ubi8 && docker buildx build --platform linux/amd64 -t quay.io/devfile/base-developer-image:${{ env.PR_NUMBER }} .
+       cd base/ubi8 && docker build -t quay.io/devfile/base-developer-image:ubi8-latest .
 
     - name: Build universal image
       run: |
         eval $(minikube docker-env)
-        cd universal/ubi8 && docker buildx build --platform linux/amd64 -t quay.io/devfile/base-developer-image:${{ env.PR_NUMBER }} .
+        cd universal/ubi8 && docker build -t quay.io/devfile/base-developer-image:${{ env.PR_NUMBER }} .
 
     - name: Checkout DWO
       uses: actions/checkout@master
--- a/.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
+++ b/.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
@@ -58,7 +58,7 @@ jobs:
     - name: Build universal image
       run: |
         eval $(minikube docker-env)
-        cd universal/ubi8 && docker build -t quay.io/devfile/base-developer-image:${{ env.PR_NUMBER }} .
+        cd universal/ubi8 && docker build -t quay.io/devfile/universal-developer-image:${{ env.PR_NUMBER }} .
 
     - name: Checkout DWO
       uses: actions/checkout@master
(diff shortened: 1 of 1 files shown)


.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
```diff
@@ -19,10 +19,12 @@ on:
       - LICENSE
       - '.rebase/*'
 
+
```
l0rd, line 22:
```suggestion
```

.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
```diff
@@ -32,57 +34,39 @@ jobs:
...
+        minikube-version: v1.31.0
 
+     # connect with docker daemon in the minikube and build an image there
+     # we need to build the image in the minikube because we have just 14 GB of space on the runner
+     # the UBI have more than 9 GB size this approach saves the disk space
     - name: Build base image
       run: |
-        cd base/ubi8 && docker buildx build --platform linux/amd64 -t quay.io/devfile/base-developer-image:${{ env.PR_NUMBER }} .
+       eval $(minikube docker-env)
+       cd base/ubi8 && docker buildx build --platform linux/amd64 -t quay.io/devfile/base-developer-image:${{ env.PR_NUMBER }} .
```
l0rd, line 57: Do you need the `buildx` argument? Isn't it the default?

.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
```diff
@@ -32,57 +34,39 @@ jobs:
...
-      run: |
-        docker push quay.io/devfile/universal-developer-image:${{ env.PR_NUMBER }}
-
-    - name: Start minikube cluster
-      id: run-minikube
-      uses: che-incubator/setup-minikube-action@next
-      with:
-        minikube-version: v1.30.1
+        eval $(minikube docker-env)
+        cd universal/ubi8 && docker buildx build --platform linux/amd64 -t quay.io/devfile/base-developer-image:${{ env.PR_NUMBER }} .
```
l0rd, line 62: You are tagging this base image with the PR number. But the build of the universal image will use `ubi8-latest` as the base image, not the one that you just built ([FROM quay.io/devfile/base-developer-image:ubi8-latest](https://github.com/devfile/developer-images/blob/main/universal/ubi8/Dockerfile#L4C43-L4C54)). You should tag the base image with `ubi8-latest` too.

l0rd (changes requested):

.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
musienko-maxim: Is it make sense to remove this job in this case? I guess this image can be pulled from the registry and we do not need to build locally.

l0rd: No. The PR can be about a change to the base image so we need to build both no?

musienko-maxim: Ok, I just changed it on ubi8-latest. The build base image is here.

.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
```diff
@@ -100,10 +83,23 @@ jobs:
...
+      with:
+        ref: api-test-with-clone-project-without-generating
+        repository: eclipse/che
+        path: che
+
     - name: Run Empty workspace smoke test
       run: |
-        export TS_API_TEST_UDI_IMAGE=quay.io/devfile/universal-developer-image:${{ env.PR_NUMBER }}
-        export TS_API_TEST_CHE_CODE_EDITOR_DEVFILE_URI=https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/che-incubator/che-code/latest/devfile.yaml
+        export TS_API_TEST_UDI_IMAGE=quay.io/devfile/base-developer-image:${{ env.PR_NUMBER }}
```
l0rd, line 102: maybe 

```suggestion
        export TS_API_TEST_UDI_IMAGE=quay.io/devfile/universal-developer-image:${{ env.PR_NUMBER }}
```

.github/workflows/empty-worksapce-smoke-test-on-minikube-ubi8.yaml
```diff
@@ -32,57 +33,39 @@ jobs:
...
-      run: |
-        docker push quay.io/devfile/universal-developer-image:${{ env.PR_NUMBER }}
-
-    - name: Start minikube cluster
-      id: run-minikube
-      uses: che-incubator/setup-minikube-action@next
-      with:
-        minikube-version: v1.30.1
+        eval $(minikube docker-env)
+        cd universal/ubi8 && docker build -t quay.io/devfile/base-developer-image:${{ env.PR_NUMBER }} .
```
l0rd, line 61:
```suggestion
        cd universal/ubi8 && docker build -t quay.io/devfile/universal-developer-image:${{ env.PR_NUMBER }} .
```

l0rd (approved):
👍

PR #113, opened 2023-08-16
Don't manually add Node to PATH

AObuchow:
Removes the manual modification of PATH in the Dockerfile as running `nvm use v$NODEJS_VERSION` accomplishes this in a more reliable way. 

I've pushed `quay.io/aobuchow/universal-developer-image:no-node-on-path` to test the changes from this PR.
To verify Node still works as expected, run `node --version` and ensure the output is still `v16.14.0`.

Additionally, here's a comparison of PATH before and after this change.
Before:

```bash
$ echo $PATH | tr : \\n | sort
/bin
/home/user/.cargo/bin
/home/user/.dotnet/tools
/home/user/.krew/bin
/home/user/.krew/bin
/home/user/.local/bin
/home/user/.local/share/coursier/bin
/home/user/.nvm/versions/node/v16.14.0/bin
/home/user/.sdkman/candidates/gradle/current/bin
/home/user/.sdkman/candidates/java/current/bin
/home/user/.sdkman/candidates/jbang/current/bin
/home/user/.sdkman/candidates/maven/current/bin
/home/user/bin
/home/user/go/bin/
/sbin
/usr/bin
/usr/local/bin
/usr/local/sbin
/usr/sbin
/usr/share/Modules/bin
```

After:

```
$ echo $PATH | tr : \\n | sort
/bin
/home/user/.cargo/bin
/home/user/.dotnet/tools
/home/user/.krew/bin
/home/user/.krew/bin
/home/user/.local/bin
/home/user/.local/share/coursier/bin
/home/user/.nvm/versions/node/v16.14.0/bin
/home/user/.sdkman/candidates/gradle/current/bin
/home/user/.sdkman/candidates/java/current/bin
/home/user/.sdkman/candidates/jbang/current/bin
/home/user/.sdkman/candidates/maven/current/bin
/home/user/bin
/home/user/go/bin/
/sbin
/usr/bin
/usr/local/bin
/usr/local/sbin
/usr/sbin
/usr/share/Modules/bin
```

Side note: [Krew is being added to PATH twice](https://github.com/devfile/developer-images/blob/main/universal/ubi8/Dockerfile#L269-L270). I'll create an issue for this.  

Fix eclipse/che#22436

PR #110, opened 2023-08-03
Organize bash completions

AObuchow:
This PR aims to resolve eclipse/che#22411.

## What does this PR change?
The following changes are made:
1. The `rcupdate=false` request parameter is used when installing sdkman to prevent the `.bashrc`  from being automatically modified. 
2. The sdkman init script launcher is now being manually added to the end of the `.bashrc`. In my testing, running the `sdk` command still worked prior to this change, however, having the sdkman init script at the end of the `.bashrc` is recommended.
3. There is now a section dedicated to bash completions at the end of the UDI dockerfile. 
4. The `bash-completion` package was added and `/etc/profile.d/bash_completion.sh` is being sourced in the  `.bashrc`, so that all completions stored in `/usr/share/bash-completion/completions/` are now active.  This has a side-effect of installing additional completions that were previously not enabled, such as `flock`.
5. The nvm installer is no longer automatically modifying the `.bashrc`. Instead, we now manually source `nvm.sh` in the `.bashrc`, and add the nvm completions to `/usr/share/bash-completion/completions/`

This PR could be improved to remove direct reference to `/usr/share/bash-completion/completions/` and instead use pkg-config to obtain this location, [as is done in WTO](https://github.com/redhat-developer/web-terminal-tooling/blob/b0ffeb64ee771c1efd91acd3ff9efd7d27173985/Dockerfile#L29). I opted not to do this to prevent bringing in pkg-config as an additional dependency, however I'd be happy to change this.

Prior to this PR, the `/home/user/.bashrc` resembled the following:

```bash
# .bashrc

# Source global definitions
if [ -f /etc/bashrc ]; then
        . /etc/bashrc
fi

# User specific environment
if ! [[ "$PATH" =~ "$HOME/.local/bin:$HOME/bin:" ]]
then
    PATH="$HOME/.local/bin:$HOME/bin:$PATH"
fi
export PATH

# Uncomment the following line if you don't like systemctl's auto-paging feature:
# export SYSTEMD_PAGER=

# User specific aliases and functions
export PS1='\W `git branch --show-current 2>/dev/null | sed -r -e "s@^(.+)@\(\1\) @"`$ '

#THIS MUST BE AT THE END OF THE FILE FOR SDKMAN TO WORK!!!
export SDKMAN_DIR="$HOME/.sdkman"
[[ -s "$HOME/.sdkman/bin/sdkman-init.sh" ]] && source "$HOME/.sdkman/bin/sdkman-init.sh"

[...]

PR #115, opened 2023-08-29
Move tools & configs from /home/user/ to /home/tooling/

AObuchow:
This PR aims to resolve https://github.com/eclipse/che/issues/22412, where enabling `devEnvironments.persistUserHome` in the Che Cluster CR will cause the workspace's PVC to mount to `/home/user/`, overwriting any tools, configurations, etc. that are present in the UDI's  `/home/user/` directory.

In order to resolve this issue, we are now installing tools (and most config files) to `/home/tooling/` and using GNU stow to create symbolic links to `/home/user/` for all files in `/home/tooling/` (with the exception of `.viminfo`, which cannot be a symbolic link by vim's design). 

Rather than trying to keep `$PATH` to be exactly the same as prior to this change (i.e. tooling binaries are found in `/home/user/`) , `$PATH` has been modified to contain the appropriate directories in `/home/tooling/`. The reasoning for this change is due to the fact that the call to stow in the entrypoint is [not guaranteed](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks) to run before any postStart commands. In order to ensure postStart commands can invoke all UDI binaries, they must be placed under `/home/tooling/`, which guarantees they will not be overwritten when a PVC is mounted to /home/user/.

Below are some other details regarding the changes made in this PR:

1. Stow is invoked with the  `--no-folding` option to recreate the directory tree structure from `/home/tooling/` in `/home/user/`, to ensure any changes within a configuration folder will persist on the PVC (as directories in `/home/tooling/` are essentially ephemeral). 
2. Stow is invoked up to 3 times: 
   - Once in the UBI8 base image, where the `--no-folding` option is not used due to the lack of directories in the /home/tooling/ of the base image.
   - Once in the UDI, with `--no-folding` enabled.
   - Once in the UDI's entrypoint, if we detect that `/home/user/` has been mounted by a PVC (erasing the contents of `/home/user/`)
3. Stow does _not_ create a symbolic link for `.viminfo`, as it [must be a regular file by vim's design](https://github.com/vim/vim/issues/2506#issuecomment-1154341891). Consequently, the `.viminfo` must be manually copied from `/home/tooling/` -> `/home/user/`.
4. Stow adds a bit of extra startup time to the container when it is invoked in the UDI's entrypoint . Unfortunately, we cannot run it as a background process or else it [might not complete successfully when the container is started](https://github.com/devfile/developer-images/pull/115#issuecomment-1714761293).
5. The changes from https://github.com/devfile/developer-images/pull/117 were adopted for this PR. I kept `/home/user/.local/bin/` still on `$PATH` so that users could mount their own binaries to `/home/user/.local/bin/` and invoke them as VSCode tasks or postStart commands. 
6. `$SDKMAN_DIR` and `$NVM_DIR` are hard-coded to point to `/home/tooling/.sdkman` (rather than `$HOME/.sdkman`). sdkman will not function if `$SDKMAN_DIR` is set to `/home/user/.sdkman/`, as its init script [won't detect symlinks](https://github.com/sdkman/sdkman-cli/blob/master/src/main/bash/sdkman-init.sh#L75).
7. We are no longer modifying the `~/.bashrc` to extend the bash environment. Instead, we now have 2 bash scripts located in `/etc/profile.d/`: one which modifies the bash environment, another to set the terminal prompt. These scripts are automatically loaded by bash. Now, users are free to modify the UDI's .bashrc (e.g. overwritting it with an automount configmap in Che) without breaking any modifications to `$PATH` that we provide in the UDI (such as sdkman).
8. I removed code from the UBI8's entrypoint that involved manually creating a `~/.bashrc`, [which was obsolete](https://github.com/devfile/developer-images/pull/115#issuecomment-1726996106).
9. The entrypoint now looks for the kubeconfig in `$KUBECONFIG` rather than being hard-coded to `/home/user/.kube/config`.
10. The kubedock entrypoint was modified to no longer fail if the kube config could not be found.
11. The duration of the timeout for finding the kube config in the entrypoint can now be changed with `$KUBEDOCK_TIMEOUT`

This PR is also based off of https://github.com/devfile/developer-images/pull/110, so reading the changes introduced in that PR is also worthwhile. 

## Testing
I've pushed `quay.io/aobuchow/universal-developer-image:stow` to test the UDI as a docker container. I've also created a sample [devfile](https://github.com/AObuchow/udi-tooling/blob/master/devfile.yaml) that uses the updated UDI, to test in Che with `devEnvironments.persistUserHome.enabled: true`.

### Sample devfile notes
My sample devfile [invokes some commands](https://github.com/AObuchow/udi-tooling/blob/master/devfile.yaml#L63) to ensure they are available on $PATH to run as preStart commands. Note: some of these commands generate stderr output, which can be viewed in `/tmp/poststart-stderr.txt`.

Additionally, I attempted to [log the binaries available as postStart commands](https://github.com/AObuchow/udi-tooling/blob/master/devfile.yaml#L65-L69) with `compgen -c | xargs which --all --skip-functions | sort | uniq > /projects/udi-tooling/binaries.txt ; ` [Here's the output of that command](https://github.com/AObuchow/udi-tooling/blob/master/binaries.txt).

One final note about my devfile: since postStart commands in DWO are run as `/bin/sh -C { <commands> }` any modifications to `$PATH` present in the bash environment are normally not usable (e.g. nvm cannot be invoked). To workaround this, I added a postStart command that runs source `/home/user/.bashrc` in the parent devfile. This workaround could potentially be used in CheCode when running VSCode tasks ([relevant comment](https://github.com/devfile/developer-images/pull/113#issuecomment-1710612441)). 

### Other testing notes

Properly ensuring this PR does not break any existing functionality or workflows may be tricky due to the nature of moving everything out of `/home/user/` into `/home/tooling/` (and then "back" into `/home/user/` with stow).

Since the podman wrapper was affected by this change, it's also worth ensuring the test cases ([wrapped podman commands](https://github.com/devfile/developer-images/pull/107/files#diff-22935aa66ea1760572f5547f38f94faf4abd071e88a5a33f6a92e139f79cb54dR5)) from https://github.com/devfile/developer-images/pull/107 work as well. 

Below are some comparisons of the `/home/user/` directory, the `/home/user/.bashrc` and `$PATH`, before and after this change. 

#### Before:

```bash
~ $ ls -la
total 108
drwxrwx---. 1 user root   244 Aug 29 02:12 .
drwxrwxr-x. 1 root root     8 Jul 17 19:51 ..
-rw-rw-r--. 1 user root    18 Jun 20  2022 .bash_logout
-rw-rw-r--. 1 user root   141 Jun 20  2022 .bash_profile
-rw-rw-r--. 1 user root   826 Aug  3 17:34 .bashrc
drwxrwxr-x. 1 root root    22 Aug  3 17:30 .cache
drwxrwxr-x. 1 root root    12 Aug  3 17:31 .cargo
drwxrwxr-x. 1 root root    36 Aug  3 17:33 .config
-rw-rw-r--. 1 root root   126 Jul 17 19:51 .gitconfig
drwxrwxr-x. 1 root root    42 Aug  3 17:33 .krew
-rw-rw-r--. 1 root root 88043 Aug  3 17:33 .kubectl_aliases
drwxr-xr-x. 1 user user     6 Aug 29 02:12 .local
drwxrwxr-x. 1 user root    84 Aug  3 17:27 .npm
drwxrwxr-x. 1 user root   610 Aug  3 17:27 .nvm
drwxrwxr-x. 1 root root    96 Aug  3 17:31 .rustup
drwxrwxr-x. 1 user root    84 Aug  3 04:47 .sdkman
-rw-rw-r--. 1 root root   172 Aug  3 17:28 .wget-hsts
drwxrwxr-x. 1 root root    12 Aug  3 17:30 go
```
#### After:
```bash
~ $ ls -la
total 28
drwxrwx---. 1 user root 260 Sep 20 04:28 .
drwxrwxr-x. 1 root root  22 Sep 20 04:07 ..
-rw-rw-r--. 1 user root  18 Jun 20  2022 .bash_logout
-rw-rw-r--. 1 user root 141 Jun 20  2022 .bash_profile
-rw-rw-r--. 1 user root 376 Jun 20  2022 .bashrc
drwxrwxr-x. 1 root root  22 Sep 20 04:28 .cache
drwxrwxr-x. 1 root root  12 Sep 20 04:28 .cargo
drwxrwxr-x. 1 root root  36 Sep 20 04:28 .config
lrwxrwxrwx. 1 root root  21 Sep 20 04:07 .gitconfig -> ../tooling/.gitconfig
drwxrwxr-x. 1 root root  42 Sep 20 04:28 .krew
lrwxrwxrwx. 1 root root  27 Sep 20 04:28 .kubectl_aliases -> ../tooling/.kubectl_aliases
drwxrwxr-x. 1 root root   6 Sep 20 04:28 .local
drwxrwxr-x. 1 root root  84 Sep 20 04:28 .npm
drwxrwxr-x. 1 root root 610 Sep 20 04:28 .nvm
drwxrwxr-x. 1 root root  96 Sep 20 04:28 .rustup
drwxrwxr-x. 1 root root  84 Sep 20 04:28 .sdkman
-rw-rw----. 1 root root 532 Sep 20 04:07 .viminfo
lrwxrwxrwx. 1 root root  21 Sep 20 04:28 .wget-hsts -> ../tooling/.wget-hsts
drwxrwxr-x. 1 root root  12 Sep 20 04:28 go
```
### `/home/user/` directory tree

Directory tree created using https://github.com/a8m/tree/tree/master: installed by running `go install github.com/a8m/tree/cmd/tree@latest` then `go clean -cache`

#### Before:
```bash
.
├── .bash_logout
├── .bash_profile
├── .bashrc
├── .cache
│   ├── go-build
│   │   ├── README
│   │   └── trim.txt
│   └── pip
│       ├── http
│       └── selfcheck
├── .cargo
│   ├── bin
│   │   ├── cargo
│   │   ├── cargo-clippy
│   │   ├── cargo-fmt
│   │   ├── cargo-miri
│   │   ├── clippy-driver
│   │   ├── rls
│   │   ├── rust-analyzer
│   │   ├── rust-gdb
│   │   ├── rust-gdbgui
│   │   ├── rust-lldb
│   │   ├── rustc
│   │   ├── rustdoc
│   │   ├── rustfmt
│   │   └── rustup
│   └── env
├── .config
│   ├── composer
│   │   ├── keys.dev.pub
│   │   └── keys.tags.pub
│   └── containers
│       └── storage.conf
├── .gitconfig
├── .krew
│   ├── bin
│   │   ├── kubectl-ctx -> /home/user/.krew/store/ctx/v0.9.5/kubectx
│   │   ├── kubectl-krew -> /home/user/.krew/store/krew/v0.4.4/krew
│   │   └── kubectl-ns -> /home/user/.krew/store/ns/v0.9.5/kubens
│   ├── index
│   │   └── default
│   ├── receipts
│   │   ├── ctx.yaml
│   │   ├── krew.yaml
│   │   └── ns.yaml
│   └── store
│       ├── ctx
│       ├── krew
│       └── ns
├── .kubectl_aliases
├── .npm
│   ├── _cacache
│   │   ├── content-v2
│   │   ├── index-v5
│   │   └── tmp
│   ├── _logs
│   │   ├── 2023-07-17T19_53_43_861Z-debug-0.log
│   │   ├── 2023-07-17T19_53_48_751Z-debug-0.log
│   │   ├── 2023-07-17T19_53_49_051Z-debug-0.log
│   │   └── 2023-07-17T19_53_49_684Z-debug-0.log
│   └── _update-notifier-last-checked
├── .nvm
│   ├── .cache
│   │   └── bin
│   ├── .dockerignore
│   ├── .editorconfig
│   ├── .git
│   │   ├── FETCH_HEAD
│   │   ├── HEAD
│   │   ├── branches
│   │   ├── config
│   │   ├── description
│   │   ├── hooks
│   │   ├── index
│   │   ├── info
│   │   ├── logs
│   │   ├── objects
│   │   ├── packed-refs
│   │   ├── refs
│   │   └── shallow
│   ├── .gitattributes
│   ├── .github
│   │   ├── FUNDING.yml
│   │   ├── ISSUE_TEMPLATE.md
│   │   ├── SECURITY.md
│   │   └── workflows
│   ├── .gitignore
│   ├── .mailmap
│   ├── .npmrc
│   ├── .travis.yml
│   ├── CODE_OF_CONDUCT.md
│   ├── CONTRIBUTING.md
│   ├── Dockerfile
│   ├── GOVERNANCE.md
│   ├── LICENSE.md
│   ├── Makefile
│   ├── PROJECT_CHARTER.md
│   ├── README.md
│   ├── ROADMAP.md
│   ├── alias
│   │   ├── default
│   │   └── lts
│   ├── bash_completion
│   ├── install.sh
│   ├── nvm-exec
│   ├── nvm.sh
│   ├── package.json
│   ├── rename_test.sh
│   ├── test
│   │   ├── common.sh
│   │   ├── fast
│   │   ├── install_script
│   │   ├── installation_iojs
│   │   ├── installation_node
│   │   ├── mocks
│   │   ├── slow
│   │   └── sourcing
│   ├── update_test_mocks.sh
│   └── versions
│       └── node
├── .rustup
│   ├── downloads
│   ├── settings.toml
│   ├── tmp
│   ├── toolchains
│   │   └── stable-x86_64-unknown-linux-gnu
│   └── update-hashes
│       └── stable-x86_64-unknown-linux-gnu
├── .sdkman
│   ├── bin
│   │   └── sdkman-init.sh
│   ├── candidates
│   │   ├── gradle
│   │   ├── java
│   │   ├── jbang
│   │   └── maven
│   ├── contrib
│   │   └── completion
│   ├── etc
│   │   └── config
│   ├── ext
│   ├── libexec
│   │   ├── default
│   │   ├── help
│   │   ├── home
│   │   ├── uninstall
│   │   └── version
│   ├── src
│   │   ├── sdkman-availability.sh
│   │   ├── sdkman-cache.sh
│   │   ├── sdkman-config.sh
│   │   ├── sdkman-current.sh
│   │   ├── sdkman-default.sh
│   │   ├── sdkman-env-helpers.sh
│   │   ├── sdkman-env.sh
│   │   ├── sdkman-flush.sh
│   │   ├── sdkman-help.sh
│   │   ├── sdkman-home.sh
│   │   ├── sdkman-install.sh
│   │   ├── sdkman-list.sh
│   │   ├── sdkman-main.sh
│   │   ├── sdkman-offline.sh
│   │   ├── sdkman-path-helpers.sh
│   │   ├── sdkman-selfupdate.sh
│   │   ├── sdkman-uninstall.sh
│   │   ├── sdkman-update.sh
│   │   ├── sdkman-upgrade.sh
│   │   ├── sdkman-use.sh
│   │   ├── sdkman-utils.sh
│   │   └── sdkman-version.sh
│   ├── tmp
│   └── var
│       ├── candidates
│       ├── delay_upgrade
│       ├── metadata
│       ├── platform
│       ├── version
│       └── version_native
├── .wget-hsts
├── .zshrc
└── go
    ├── bin
    │   ├── gopls
    │   └── tree
    └── pkg
        ├── mod
        └── sumdb

76 directories, 111 files
```

#### After:

```bash
~ $ tree -l -a
.
├── .bash_logout
├── .bash_profile
├── .bashrc
├── .cache
│   ├── go-build
│   │   ├── README -> ../../../tooling/.cache/go-build/README
│   │   └── trim.txt -> ../../../tooling/.cache/go-build/trim.txt
│   └── pip
│       ├── http
│       └── selfcheck
├── .cargo
│   ├── bin
│   │   ├── cargo -> ../../../tooling/.cargo/bin/cargo
│   │   ├── cargo-clippy -> ../../../tooling/.cargo/bin/cargo-clippy
│   │   ├── cargo-fmt -> ../../../tooling/.cargo/bin/cargo-fmt
│   │   ├── cargo-miri -> ../../../tooling/.cargo/bin/cargo-miri
│   │   ├── clippy-driver -> ../../../tooling/.cargo/bin/clippy-driver
│   │   ├── rls -> ../../../tooling/.cargo/bin/rls
│   │   ├── rust-analyzer -> ../../../tooling/.cargo/bin/rust-analyzer
│   │   ├── rust-gdb -> ../../../tooling/.cargo/bin/rust-gdb
│   │   ├── rust-gdbgui -> ../../../tooling/.cargo/bin/rust-gdbgui
│   │   ├── rust-lldb -> ../../../tooling/.cargo/bin/rust-lldb
│   │   ├── rustc -> ../../../tooling/.cargo/bin/rustc
│   │   ├── rustdoc -> ../../../tooling/.cargo/bin/rustdoc
│   │   ├── rustfmt -> ../../../tooling/.cargo/bin/rustfmt
│   │   └── rustup -> ../../../tooling/.cargo/bin/rustup
│   └── env -> ../../tooling/.cargo/env
├── .config
│   ├── composer
│   │   ├── keys.dev.pub -> ../../../tooling/.config/composer/keys.dev.pub
│   │   └── keys.tags.pub -> ../../../tooling/.config/composer/keys.tags.pub
│   └── containers
│       └── storage.conf -> ../../../tooling/.config/containers/storage.conf
├── .gitconfig -> ../tooling/.gitconfig
├── .krew
│   ├── bin
│   │   ├── kubectl-ctx -> ../../../tooling/.krew/bin/kubectl-ctx
│   │   ├── kubectl-krew -> ../../../tooling/.krew/bin/kubectl-krew
│   │   └── kubectl-ns -> ../../../tooling/.krew/bin/kubectl-ns
│   ├── index
│   │   └── default
│   ├── receipts
│   │   ├── ctx.yaml -> ../../../tooling/.krew/receipts/ctx.yaml
│   │   ├── krew.yaml -> ../../../tooling/.krew/receipts/krew.yaml
│   │   └── ns.yaml -> ../../../tooling/.krew/receipts/ns.yaml
│   └── store
│       ├── ctx
│       ├── krew
│       └── ns
├── .kubectl_aliases -> ../tooling/.kubectl_aliases
├── .local
│   └── bin
├── .npm
│   ├── _cacache
│   │   ├── content-v2
│   │   ├── index-v5
│   │   └── tmp
│   ├── _logs
│   │   ├── 2023-09-20T04_14_47_221Z-debug-0.log -> ../../../tooling/.npm/_logs/2023-09-20T04_14_47_221Z-debug-0.log
│   │   ├── 2023-09-20T04_15_31_306Z-debug-0.log -> ../../../tooling/.npm/_logs/2023-09-20T04_15_31_306Z-debug-0.log
│   │   ├── 2023-09-20T04_15_31_572Z-debug-0.log -> ../../../tooling/.npm/_logs/2023-09-20T04_15_31_572Z-debug-0.log
│   │   └── 2023-09-20T04_15_38_586Z-debug-0.log -> ../../../tooling/.npm/_logs/2023-09-20T04_15_38_586Z-debug-0.log
│   └── _update-notifier-last-checked -> ../../tooling/.npm/_update-notifier-last-checked
├── .nvm
│   ├── .cache
│   │   └── bin
│   ├── .dockerignore -> ../../tooling/.nvm/.dockerignore
│   ├── .editorconfig -> ../../tooling/.nvm/.editorconfig
│   ├── .git
│   │   ├── FETCH_HEAD -> ../../../tooling/.nvm/.git/FETCH_HEAD
│   │   ├── HEAD -> ../../../tooling/.nvm/.git/HEAD
│   │   ├── branches
│   │   ├── config -> ../../../tooling/.nvm/.git/config
│   │   ├── description -> ../../../tooling/.nvm/.git/description
│   │   ├── hooks
│   │   ├── index -> ../../../tooling/.nvm/.git/index
│   │   ├── info
│   │   ├── logs
│   │   ├── objects
│   │   ├── packed-refs -> ../../../tooling/.nvm/.git/packed-refs
│   │   ├── refs
│   │   └── shallow -> ../../../tooling/.nvm/.git/shallow
│   ├── .gitattributes -> ../../tooling/.nvm/.gitattributes
│   ├── .github
│   │   ├── FUNDING.yml -> ../../../tooling/.nvm/.github/FUNDING.yml
│   │   ├── ISSUE_TEMPLATE.md -> ../../../tooling/.nvm/.github/ISSUE_TEMPLATE.md
│   │   ├── SECURITY.md -> ../../../tooling/.nvm/.github/SECURITY.md
│   │   └── workflows
│   ├── .gitignore -> ../../tooling/.nvm/.gitignore
│   ├── .mailmap -> ../../tooling/.nvm/.mailmap
│   ├── .npmrc -> ../../tooling/.nvm/.npmrc
│   ├── .travis.yml -> ../../tooling/.nvm/.travis.yml
│   ├── CODE_OF_CONDUCT.md -> ../../tooling/.nvm/CODE_OF_CONDUCT.md
│   ├── CONTRIBUTING.md -> ../../tooling/.nvm/CONTRIBUTING.md
│   ├── Dockerfile -> ../../tooling/.nvm/Dockerfile
│   ├── GOVERNANCE.md -> ../../tooling/.nvm/GOVERNANCE.md
│   ├── LICENSE.md -> ../../tooling/.nvm/LICENSE.md
│   ├── Makefile -> ../../tooling/.nvm/Makefile
│   ├── PROJECT_CHARTER.md -> ../../tooling/.nvm/PROJECT_CHARTER.md
│   ├── README.md -> ../../tooling/.nvm/README.md
│   ├── ROADMAP.md -> ../../tooling/.nvm/ROADMAP.md
│   ├── alias
│   │   ├── default -> ../../../tooling/.nvm/alias/default
│   │   └── lts
│   ├── bash_completion -> ../../tooling/.nvm/bash_completion
│   ├── install.sh -> ../../tooling/.nvm/install.sh
│   ├── nvm-exec -> ../../tooling/.nvm/nvm-exec
│   ├── nvm.sh -> ../../tooling/.nvm/nvm.sh
│   ├── package.json -> ../../tooling/.nvm/package.json
│   ├── rename_test.sh -> ../../tooling/.nvm/rename_test.sh
│   ├── test
│   │   ├── common.sh -> ../../../tooling/.nvm/test/common.sh
│   │   ├── fast
│   │   ├── install_script
│   │   ├── installation_iojs
│   │   ├── installation_node
│   │   ├── mocks
│   │   ├── slow
│   │   └── sourcing
│   ├── update_test_mocks.sh -> ../../tooling/.nvm/update_test_mocks.sh
│   └── versions
│       └── node
├── .rustup
│   ├── downloads
│   ├── settings.toml -> ../../tooling/.rustup/settings.toml
│   ├── tmp
│   ├── toolchains
│   │   └── stable-x86_64-unknown-linux-gnu
│   └── update-hashes
│       └── stable-x86_64-unknown-linux-gnu -> ../../../tooling/.rustup/update-hashes/stable-x86_64-unknown-linux-gnu
├── .sdkman
│   ├── bin
│   │   └── sdkman-init.sh -> ../../../tooling/.sdkman/bin/sdkman-init.sh
│   ├── candidates
│   │   ├── gradle
│   │   ├── java
│   │   ├── jbang
│   │   └── maven
│   ├── contrib
│   │   └── completion
│   ├── etc
│   │   └── config -> ../../../tooling/.sdkman/etc/config
│   ├── ext
│   ├── libexec
│   │   ├── default -> ../../../tooling/.sdkman/libexec/default
│   │   ├── help -> ../../../tooling/.sdkman/libexec/help
│   │   ├── home -> ../../../tooling/.sdkman/libexec/home
│   │   ├── uninstall -> ../../../tooling/.sdkman/libexec/uninstall
│   │   └── version -> ../../../tooling/.sdkman/libexec/version
│   ├── src
│   │   ├── sdkman-availability.sh -> ../../../tooling/.sdkman/src/sdkman-availability.sh
│   │   ├── sdkman-cache.sh -> ../../../tooling/.sdkman/src/sdkman-cache.sh
│   │   ├── sdkman-config.sh -> ../../../tooling/.sdkman/src/sdkman-config.sh
│   │   ├── sdkman-current.sh -> ../../../tooling/.sdkman/src/sdkman-current.sh
│   │   ├── sdkman-default.sh -> ../../../tooling/.sdkman/src/sdkman-default.sh
│   │   ├── sdkman-env-helpers.sh -> ../../../tooling/.sdkman/src/sdkman-env-helpers.sh
│   │   ├── sdkman-env.sh -> ../../../tooling/.sdkman/src/sdkman-env.sh
│   │   ├── sdkman-flush.sh -> ../../../tooling/.sdkman/src/sdkman-flush.sh
│   │   ├── sdkman-help.sh -> ../../../tooling/.sdkman/src/sdkman-help.sh
│   │   ├── sdkman-home.sh -> ../../../tooling/.sdkman/src/sdkman-home.sh
│   │   ├── sdkman-install.sh -> ../../../tooling/.sdkman/src/sdkman-install.sh
│   │   ├── sdkman-list.sh -> ../../../tooling/.sdkman/src/sdkman-list.sh
│   │   ├── sdkman-main.sh -> ../../../tooling/.sdkman/src/sdkman-main.sh
│   │   ├── sdkman-offline.sh -> ../../../tooling/.sdkman/src/sdkman-offline.sh
│   │   ├── sdkman-path-helpers.sh -> ../../../tooling/.sdkman/src/sdkman-path-helpers.sh
│   │   ├── sdkman-selfupdate.sh -> ../../../tooling/.sdkman/src/sdkman-selfupdate.sh
│   │   ├── sdkman-uninstall.sh -> ../../../tooling/.sdkman/src/sdkman-uninstall.sh
│   │   ├── sdkman-update.sh -> ../../../tooling/.sdkman/src/sdkman-update.sh
│   │   ├── sdkman-upgrade.sh -> ../../../tooling/.sdkman/src/sdkman-upgrade.sh
│   │   ├── sdkman-use.sh -> ../../../tooling/.sdkman/src/sdkman-use.sh
│   │   ├── sdkman-utils.sh -> ../../../tooling/.sdkman/src/sdkman-utils.sh
│   │   └── sdkman-version.sh -> ../../../tooling/.sdkman/src/sdkman-version.sh
│   ├── tmp
│   └── var
│       ├── candidates -> ../../../tooling/.sdkman/var/candidates
│       ├── delay_upgrade -> ../../../tooling/.sdkman/var/delay_upgrade
│       ├── metadata
│       ├── platform -> ../../../tooling/.sdkman/var/platform
│       ├── version -> ../../../tooling/.sdkman/var/version
│       └── version_native -> ../../../tooling/.sdkman/var/version_native
├── .viminfo
├── .wget-hsts -> ../tooling/.wget-hsts
└── go
    ├── bin
    │   └── gopls -> ../../../tooling/go/bin/gopls
    └── pkg
        ├── mod
        └── sumdb

78 directories, 110 files
```

#### Before:

```bash
~ $ cat /home/user/.bashrc
# .bashrc

# Source global definitions
if [ -f /etc/bashrc ]; then
        . /etc/bashrc
fi

# User specific environment
if ! [[ "$PATH" =~ "$HOME/.local/bin:$HOME/bin:" ]]
then
    PATH="$HOME/.local/bin:$HOME/bin:$PATH"
fi
export PATH

# Uncomment the following line if you don't like systemctl's auto-paging feature:
# export SYSTEMD_PAGER=

# User specific aliases and functions
export PS1='\W `git branch --show-current 2>/dev/null | sed -r -e "s@^(.+)@\(\1\) @"`$ '

#THIS MUST BE AT THE END OF THE FILE FOR SDKMAN TO WORK!!!
export SDKMAN_DIR="$HOME/.sdkman"
[[ -s "$HOME/.sdkman/bin/sdkman-init.sh" ]] && source "$HOME/.sdkman/bin/sdkman-init.sh"

export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
[ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
source /usr/share/bash-completion/completions/git
source /usr/share/bash-completion/completions/oc
[ -f ~/.kubectl_aliases ] && source ~/.kubectl_aliases
export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
```

#### After:

```bash
~ $ cat /home/user/.bashrc
# .bashrc

# Source global definitions
if [ -f /etc/bashrc ]; then
        . /etc/bashrc
fi

# User specific environment
if ! [[ "$PATH" =~ "$HOME/.local/bin:$HOME/bin:" ]]
then
    PATH="$HOME/.local/bin:$HOME/bin:$PATH"
fi
export PATH

# Uncomment the following line if you don't like systemctl's auto-paging feature:
# export SYSTEMD_PAGER=

# User specific aliases and functions
```

Also, here are the 2 new files in `/etc/profile.d/`:

```bash
profile.d $ cat udi_environment.sh 
export NVM_DIR="/home/tooling/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
[ -f ~/.kubectl_aliases ] && source ~/.kubectl_aliases
export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
export SDKMAN_DIR="/home/tooling/.sdkman"
[[ -s "$SDKMAN_DIR/bin/sdkman-init.sh" ]] && source "$SDKMAN_DIR/bin/sdkman-init.sh"
```

```bash
profile.d $ cat udi_prompt.sh 
export PS1='\W `git branch --show-current 2>/dev/null | sed -r -e "s@^(.+)@\(\1\) @"`$ '
```

#### Before:

```
~ $ echo $PATH | tr : \\n | sort
/bin
/home/user/.cargo/bin
/home/user/.dotnet/tools
/home/user/.krew/bin
/home/user/.krew/bin
/home/user/.local/bin
/home/user/.local/share/coursier/bin
/home/user/.nvm/versions/node/v16.14.0/bin
/home/user/.sdkman/candidates/gradle/current/bin
/home/user/.sdkman/candidates/java/current/bin
/home/user/.sdkman/candidates/jbang/current/bin
/home/user/.sdkman/candidates/maven/current/bin
/home/user/bin
/home/user/go/bin/
/sbin
/usr/bin
/usr/local/bin
/usr/local/sbin
/usr/sbin
/usr/share/Modules/bin
```

#### After:

```
~ $ echo $PATH | tr : \\n | sort
/bin
/home/tooling/.cargo/bin
/home/tooling/.krew/bin
/home/tooling/.local/bin
/home/tooling/.local/share/coursier/bin
/home/tooling/.nvm/versions/node/v16.14.0/bin
/home/tooling/.sdkman/candidates/gradle/current/bin
/home/tooling/.sdkman/candidates/java/current/bin
/home/tooling/.sdkman/candidates/jbang/current/bin
/home/tooling/.sdkman/candidates/maven/current/bin
/home/tooling/go/bin/
/home/user/.dotnet/tools
/home/user/.krew/bin
/home/user/.local/bin
/home/user/.local/bin
/home/user/bin
/sbin
/usr/bin
/usr/local/bin
/usr/local/sbin
/usr/sbin
/usr/share/Modules/bin
```

### DIFF
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -170,17 +170,12 @@ ENV KAMEL_VERSION 1.11.0
 RUN curl -L https://github.com/apache/camel-k/releases/download/v${KAMEL_VERSION}/camel-k-client-${KAMEL_VERSION}-linux-64bit.tar.gz | tar -C /usr/local/bin -xz \
     && chmod +x /usr/local/bin/kamel
 
-# git completion
-RUN echo "source /usr/share/bash-completion/completions/git" >> /home/user/.bashrc
-
 # Cloud
 
-# oc client and completion
+# oc client
 ENV OC_VERSION=4.6
 RUN curl -L https://mirror.openshift.com/pub/openshift-v4/clients/oc/${OC_VERSION}/linux/oc.tar.gz | tar -C /usr/local/bin -xz \
-    && chmod +x /usr/local/bin/oc  \
-    && oc completion bash > /usr/share/bash-completion/completions/oc \
-    && echo "source /usr/share/bash-completion/completions/oc" >> /home/user/.bashrc
+    && chmod +x /usr/local/bin/oc
 
 ## podman buildah skopeo
 RUN dnf -y module enable container-tools:rhel8 && \
@@ -237,10 +232,6 @@ dnf install -y kubectl
 curl -sSL -o ~/.kubectl_aliases https://raw.githubusercontent.com/ahmetb/kubectl-alias/master/.kubectl_aliases
 echo '[ -f ~/.kubectl_aliases ] && source ~/.kubectl_aliases' >> /home/user/.bashrc
 EOF
-
-# kubectl completion
-RUN kubectl completion bash > /usr/share/bash-completion/completions/kubectl \
-    && echo "source /usr/share/bash-completion/completions/kubectl" >> /home/user/.bashrc
 
 ## shellcheck
 RUN <<EOF
@@ -414,6 +405,16 @@ cd -
 rm -rf "${TEMP_DIR}"
 EOF
 
+# Bash completions
+RUN dnf -y install bash-completion \
+    && dnf clean all \
+    && rm -rf /var/cache/yum
+
+RUN <<EOF
+echo "source /etc/profile.d/bash_completion.sh" >> /home/user/.bashrc
+oc completion bash > /usr/share/bash-completion/completions/oc 
+kubectl completion bash > /usr/share/bash-completion/completions/kubectl
+EOF
 
 # Add sdkman's init script launcher to the end of the .bashrc since we are not adding it on sdkman install
 # NOTE: all modifications to the .bashrc must happen BEFORE this step in order for sdkman to function correctly
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -67,7 +67,10 @@ ENV NODEJS_20_VERSION=20.7.0
 # note that 18.18.0 is the latest but 18.16.1 is the supported version downstream and in ubi8
 ENV NODEJS_18_VERSION=18.16.1
 ENV NODEJS_DEFAULT_VERSION=${NODEJS_18_VERSION}
-RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.5/install.sh | bash && source /home/user/.bashrc && \
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.5/install.sh | PROFILE=/dev/null bash
+RUN echo 'export NVM_DIR="$HOME/.nvm"' >> /home/user/.bashrc \
+    && echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> /home/user/.bashrc
+RUN source /home/user/.bashrc && \
     nvm install v${NODEJS_20_VERSION} && \
     nvm install v${NODEJS_18_VERSION} && \
     nvm alias default v${NODEJS_DEFAULT_VERSION} && nvm use v${NODEJS_DEFAULT_VERSION} && \
@@ -414,6 +417,7 @@ RUN <<EOF
 echo "source /etc/profile.d/bash_completion.sh" >> /home/user/.bashrc
 oc completion bash > /usr/share/bash-completion/completions/oc 
 kubectl completion bash > /usr/share/bash-completion/completions/kubectl
+cat ${NVM_DIR}/bash_completion > /usr/share/bash-completion/completions/nvm
 EOF
 
 # Add sdkman's init script launcher to the end of the .bashrc since we are not adding it on sdkman install
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -85,15 +85,15 @@ ENV NODEJS_HOME_20=$NVM_DIR/versions/node/v${NODEJS_20_VERSION}
 ENV NODEJS_HOME_18=$NVM_DIR/versions/node/v${NODEJS_18_VERSION}
 
 # kube
-# The Che User Dashboard creates the kube config in /home/user/ and not /home/tooling/
 ENV KUBECONFIG=/home/user/.kube/config
 
 USER 0
 
 # Define user directory for binaries
-RUN mkdir -p /home/user/.local/bin && \
+RUN mkdir -p /home/tooling/.local/bin && \
     chgrp -R 0 /home && chmod -R g=u /home
 ENV PATH="/home/user/.local/bin:$PATH"
+ENV PATH="/home/tooling/.local/bin:$PATH"
 
 # Required packages for AWT
 RUN dnf install -y libXext libXrender libXtst libXi
@@ -430,12 +430,12 @@ EOF
 RUN echo 'export SDKMAN_DIR="/home/tooling/.sdkman"' >> /home/tooling/.bashrc
 RUN echo '[[ -s "$SDKMAN_DIR/bin/sdkman-init.sh" ]] && source "$SDKMAN_DIR/bin/sdkman-init.sh"' >> /home/tooling/.bashrc
 
-# Set permissions on /etc/passwd and /home to allow arbitrary users to write
-RUN chgrp -R 0 /home && chmod -R g=u /etc/passwd /etc/group /home
 
 # Create symbolic links from /home/tooling/ -> /home/user/
 RUN stow . -t /home/user/ -d /home/tooling/ --no-folding
 
+# Set permissions on /etc/passwd and /home to allow arbitrary users to write
+RUN chgrp -R 0 /home && chmod -R g=u /etc/passwd /etc/group /home
 
 # cleanup dnf cache
 RUN dnf -y clean all --enablerepo='*'
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -19,11 +19,19 @@ LABEL io.openshift.expose-services=""
 
 USER 10001
 
+# We install everything to /home/tooling/ as /home/user/ may get overriden, see github.com/eclipse/che/issues/22412
+RUN mkdir -p /home/tooling/
+ENV HOME=/home/tooling
+
+RUN cp /home/user/.bashrc /home/tooling/.bashrc
+# /home/user/.bashrc will be replaced with the a symlink from /home/tooling/.bashrc
+RUN rm /home/user/.bashrc    
+
 # Java
 RUN curl -fsSL "https://get.sdkman.io/?rcupdate=false" | bash \
-    && bash -c ". /home/user/.sdkman/bin/sdkman-init.sh \
-             && sed -i "s/sdkman_auto_answer=false/sdkman_auto_answer=true/g" /home/user/.sdkman/etc/config \
-	     && sed -i "s/sdkman_auto_env=false/sdkman_auto_env=true/g" /home/user/.sdkman/etc/config \
+    && bash -c ". /home/tooling/.sdkman/bin/sdkman-init.sh \
+             && sed -i "s/sdkman_auto_answer=false/sdkman_auto_answer=true/g" /home/tooling/.sdkman/etc/config \
+	     && sed -i "s/sdkman_auto_env=false/sdkman_auto_env=true/g" /home/tooling/.sdkman/etc/config \
              && sdk install java 8.0.332-tem \
              && sdk install java 11.0.15-tem \
              && sdk install java 17.0.3-tem \
@@ -34,53 +42,55 @@ RUN curl -fsSL "https://get.sdkman.io/?rcupdate=false" | bash \
              && sdk install jbang \
              && sdk flush archives \
              && sdk flush temp" \
-    && chgrp -R 0 /home/user && chmod -R g=u /home/user
+    && chgrp -R 0 /home/tooling && chmod -R g=u /home/tooling
 
 # sdk home java <version>
-ENV JAVA_HOME_8=/home/user/.sdkman/candidates/java/8.0.332-tem
-ENV JAVA_HOME_11=/home/user/.sdkman/candidates/java/11.0.15-tem
-ENV JAVA_HOME_17=/home/user/.sdkman/candidates/java/17.0.3-tem
+ENV JAVA_HOME_8=/home/tooling/.sdkman/candidates/java/8.0.332-tem
+ENV JAVA_HOME_11=/home/tooling/.sdkman/candidates/java/11.0.15-tem
+ENV JAVA_HOME_17=/home/tooling/.sdkman/candidates/java/17.0.3-tem
 
-# Java-related environment variables are described and set by /home/user/.bashrc
+# Java-related environment variables are described and set by /home/tooling/.bashrc
 # To make Java working for dash and other shells, it needs to initialize them in the Dockerfile.
 ENV SDKMAN_CANDIDATES_API="https://api.sdkman.io/2"
-ENV SDKMAN_CANDIDATES_DIR="/home/user/.sdkman/candidates"
-ENV SDKMAN_DIR="/home/user/.sdkman"
+ENV SDKMAN_CANDIDATES_DIR="/home/tooling/.sdkman/candidates"
+ENV SDKMAN_DIR="/home/tooling/.sdkman"
 ENV SDKMAN_PLATFORM="linuxx64"
 ENV SDKMAN_VERSION="5.13.0"
 
-ENV GRADLE_HOME="/home/user/.sdkman/candidates/gradle/current"
-ENV JAVA_HOME="/home/user/.sdkman/candidates/java/current"
-ENV MAVEN_HOME="/home/user/.sdkman/candidates/maven/current"
+ENV GRADLE_HOME="/home/tooling/.sdkman/candidates/gradle/current"
+ENV JAVA_HOME="/home/tooling/.sdkman/candidates/java/current"
+ENV MAVEN_HOME="/home/tooling/.sdkman/candidates/maven/current"
 
-ENV GRAALVM_HOME=/home/user/.sdkman/candidates/java/22.1.0.0.r17-mandrel
+ENV GRAALVM_HOME=/home/tooling/.sdkman/candidates/java/22.1.0.0.r17-mandrel
 
-ENV PATH="/home/user/.krew/bin:$PATH"
-ENV PATH="/home/user/.sdkman/candidates/maven/current/bin:$PATH"
-ENV PATH="/home/user/.sdkman/candidates/java/current/bin:$PATH"
-ENV PATH="/home/user/.sdkman/candidates/gradle/current/bin:$PATH"
-ENV PATH="/home/user/.local/share/coursier/bin:$PATH"
+ENV PATH="/home/tooling/.krew/bin:$PATH"
+ENV PATH="/home/tooling/.sdkman/candidates/maven/current/bin:$PATH"
+ENV PATH="/home/tooling/.sdkman/candidates/java/current/bin:$PATH"
+ENV PATH="/home/tooling/.sdkman/candidates/gradle/current/bin:$PATH"
+ENV PATH="/home/tooling/.local/share/coursier/bin:$PATH"
 
 # NodeJS
-ENV NVM_DIR="/home/user/.nvm"
+RUN mkdir -p /home/tooling/.nvm/
+ENV NVM_DIR="/home/tooling/.nvm"
 ENV NODEJS_20_VERSION=20.7.0
 # note that 18.18.0 is the latest but 18.16.1 is the supported version downstream and in ubi8
 ENV NODEJS_18_VERSION=18.16.1
 ENV NODEJS_DEFAULT_VERSION=${NODEJS_18_VERSION}
 RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.5/install.sh | PROFILE=/dev/null bash
-RUN echo 'export NVM_DIR="$HOME/.nvm"' >> /home/user/.bashrc \
-    && echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> /home/user/.bashrc
-RUN source /home/user/.bashrc && \
+RUN echo 'export NVM_DIR="$HOME/.nvm"' >> /home/tooling/.bashrc \
+    && echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> /home/tooling/.bashrc
+RUN source /home/tooling/.bashrc && \
     nvm install v${NODEJS_20_VERSION} && \
     nvm install v${NODEJS_18_VERSION} && \
     nvm alias default v${NODEJS_DEFAULT_VERSION} && nvm use v${NODEJS_DEFAULT_VERSION} && \
     npm install --global yarn@v1.22.17 &&\
-    chgrp -R 0 /home/user && chmod -R g=u /home/user
+    chgrp -R 0 /home/tooling && chmod -R g=u /home/tooling
 ENV PATH=$NVM_DIR/versions/node/v${NODEJS_DEFAULT_VERSION}/bin:$PATH
 ENV NODEJS_HOME_20=$NVM_DIR/versions/node/v${NODEJS_20_VERSION}
 ENV NODEJS_HOME_18=$NVM_DIR/versions/node/v${NODEJS_18_VERSION}
 
 # kube
+# The Che User Dashboard creates the kube config in /home/user/ and not /home/tooling/
 ENV KUBECONFIG=/home/user/.kube/config
 
 USER 0
@@ -112,11 +122,11 @@ RUN curl -fLo mill https://raw.githubusercontent.com/lefou/millw/main/millw && \
 RUN dnf -y install llvm-toolset gcc gcc-c++ clang clang-libs clang-tools-extra gdb
 
 # Go 1.18+    - installed to /usr/bin/go
-# gopls 0.10+ - installed to /home/user/go/bin/gopls and /home/user/go/pkg/mod/
+# gopls 0.10+ - installed to /home/tooling/go/bin/gopls and /home/tooling/go/pkg/mod/
 RUN dnf install -y go-toolset && \
     GO111MODULE=on go install -v golang.org/x/tools/gopls@latest && \
-    chgrp -R 0 /home/user && chmod -R g=u /home/user
-ENV GOBIN="/home/user/go/bin/"
+    chgrp -R 0 /home/tooling && chmod -R g=u /home/tooling
+ENV GOBIN="/home/tooling/go/bin/"
 ENV PATH="$GOBIN:$PATH"
 
 # Python
@@ -159,14 +169,14 @@ ENV DOTNET_RPM_VERSION=6.0
 RUN dnf install -y dotnet-hostfxr-${DOTNET_RPM_VERSION} dotnet-runtime-${DOTNET_RPM_VERSION} dotnet-sdk-${DOTNET_RPM_VERSION}
 
 # rust
-ENV CARGO_HOME=/home/user/.cargo \
-    RUSTUP_HOME=/home/user/.rustup \
-    PATH=/home/user/.cargo/bin:${PATH}
+ENV CARGO_HOME=/home/tooling/.cargo \
+    RUSTUP_HOME=/home/tooling/.rustup \
+    PATH=/home/tooling/.cargo/bin:${PATH}
 RUN curl --proto '=https' --tlsv1.2 -sSfo rustup https://sh.rustup.rs && \
     chmod +x rustup && \
     mv rustup /usr/bin/ && \
     rustup -y --no-modify-path --profile minimal -c rust-src -c rust-analysis -c rls && \
-    chgrp -R 0 /home/user && chmod -R g=u /home/user
+    chgrp -R 0 /home/tooling && chmod -R g=u /home/tooling
 
 # camel-k
 ENV KAMEL_VERSION 1.11.0
@@ -233,7 +243,7 @@ EOF2
 
 dnf install -y kubectl
 curl -sSL -o ~/.kubectl_aliases https://raw.githubusercontent.com/ahmetb/kubectl-alias/master/.kubectl_aliases
-echo '[ -f ~/.kubectl_aliases ] && source ~/.kubectl_aliases' >> /home/user/.bashrc
+echo '[ -f ~/.kubectl_aliases ] && source ~/.kubectl_aliases' >> /home/tooling/.bashrc
 EOF
 
 ## shellcheck
@@ -273,7 +283,7 @@ sha256sum -c "${KREW_TGZ}.sha256" 2>&1 | grep OK
 
 tar -zxvf "${KREW_TGZ}"
 ./"krew-${KREW_ARCH}" install krew
-echo 'export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"' >> /home/user/.bashrc
+echo 'export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"' >> /home/tooling/.bashrc
 export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
 # kubens and kubectx
 kubectl krew install ns
@@ -414,7 +424,7 @@ RUN dnf -y install bash-completion \
     && rm -rf /var/cache/yum
 
 RUN <<EOF
-echo "source /etc/profile.d/bash_completion.sh" >> /home/user/.bashrc
+echo "source /etc/profile.d/bash_completion.sh" >> /home/tooling/.bashrc
 oc completion bash > /usr/share/bash-completion/completions/oc 
 kubectl completion bash > /usr/share/bash-completion/completions/kubectl
 cat ${NVM_DIR}/bash_completion > /usr/share/bash-completion/completions/nvm
@@ -422,11 +432,18 @@ EOF
 
 # Add sdkman's init script launcher to the end of the .bashrc since we are not adding it on sdkman install
 # NOTE: all modifications to the .bashrc must happen BEFORE this step in order for sdkman to function correctly
-RUN echo 'export SDKMAN_DIR="/home/user/.sdkman"' >> /home/user/.bashrc
-RUN echo '[[ -s "/home/user/.sdkman/bin/sdkman-init.sh" ]] && source "/home/user/.sdkman/bin/sdkman-init.sh"' >> /home/user/.bashrc
+RUN echo 'export SDKMAN_DIR="/home/tooling/.sdkman"' >> /home/tooling/.bashrc
+RUN echo '[[ -s "$SDKMAN_DIR/bin/sdkman-init.sh" ]] && source "$SDKMAN_DIR/bin/sdkman-init.sh"' >> /home/tooling/.bashrc
 
 # Set permissions on /etc/passwd and /home to allow arbitrary users to write
 RUN chgrp -R 0 /home && chmod -R g=u /etc/passwd /etc/group /home
+
+# Create symbolic links from /home/tooling/ -> /home/user/
+RUN dnf -y install stow \
+    && dnf clean all \
+    && rm -rf /var/cache/yum \
+    &&  stow . -t /home/user/ -d /home/tooling/ --no-folding
+
 
 # cleanup dnf cache
 RUN dnf -y clean all --enablerepo='*'
@@ -434,3 +451,5 @@ RUN dnf -y clean all --enablerepo='*'
 COPY --chown=0:0 entrypoint.sh /
 
 USER 10001
+
+ENV HOME=/home/user
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -17,6 +17,12 @@ LABEL description="Image with developers tools. Languages SDK and runtimes inclu
 LABEL io.k8s.display-name="devfile-developer-universal"
 LABEL io.openshift.expose-services=""
 
+USER 0
+
+# $PROFILE_EXT contains all additions made to the bash environment
+ENV PROFILE_EXT=/etc/profile.d/udi_environment.sh
+RUN touch ${PROFILE_EXT} & chown 10001 ${PROFILE_EXT}
+
 USER 10001
 
 # We install everything to /home/tooling/ as /home/user/ may get overriden, see github.com/eclipse/che/issues/22412
@@ -44,7 +50,7 @@ ENV JAVA_HOME_8=/home/tooling/.sdkman/candidates/java/8.0.332-tem
 ENV JAVA_HOME_11=/home/tooling/.sdkman/candidates/java/11.0.15-tem
 ENV JAVA_HOME_17=/home/tooling/.sdkman/candidates/java/17.0.3-tem
 
-# Java-related environment variables are described and set by /home/tooling/.bashrc
+# Java-related environment variables are described and set by ${PROFILE_EXT}, which will be loaded by ~/.bashrc
 # To make Java working for dash and other shells, it needs to initialize them in the Dockerfile.
 ENV SDKMAN_CANDIDATES_API="https://api.sdkman.io/2"
 ENV SDKMAN_CANDIDATES_DIR="/home/tooling/.sdkman/candidates"
@@ -72,9 +78,9 @@ ENV NODEJS_20_VERSION=20.7.0
 ENV NODEJS_18_VERSION=18.16.1
 ENV NODEJS_DEFAULT_VERSION=${NODEJS_18_VERSION}
 RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.5/install.sh | PROFILE=/dev/null bash
-RUN echo 'export NVM_DIR="$HOME/.nvm"' >> /home/tooling/.bashrc \
-    && echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> /home/tooling/.bashrc
-RUN source /home/tooling/.bashrc && \
+RUN echo 'export NVM_DIR="$HOME/.nvm"' >> ${PROFILE_EXT} \
+    && echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> ${PROFILE_EXT}
+RUN source /home/user/.bashrc && \
     nvm install v${NODEJS_20_VERSION} && \
     nvm install v${NODEJS_18_VERSION} && \
     nvm alias default v${NODEJS_DEFAULT_VERSION} && nvm use v${NODEJS_DEFAULT_VERSION} && \
@@ -190,7 +196,7 @@ RUN dnf -y module enable container-tools:rhel8 && \
     dnf -y update && \
     dnf -y reinstall shadow-utils && \
     dnf -y install podman buildah skopeo fuse-overlayfs
-RUN echo 'alias docker=podman' >> /home/user/.bashrc
+RUN echo 'alias docker=podman' >> ${PROFILE_EXT}
 
 # Set up environment variables to note that this is
 # not starting with usernamespace and default to
@@ -238,7 +244,7 @@ EOF2
 
 dnf install -y kubectl
 curl -sSL -o ~/.kubectl_aliases https://raw.githubusercontent.com/ahmetb/kubectl-alias/master/.kubectl_aliases
-echo '[ -f ~/.kubectl_aliases ] && source ~/.kubectl_aliases' >> /home/tooling/.bashrc
+echo '[ -f ~/.kubectl_aliases ] && source ~/.kubectl_aliases' >> ${PROFILE_EXT}
 EOF
 
 ## shellcheck
@@ -278,7 +284,7 @@ sha256sum -c "${KREW_TGZ}.sha256" 2>&1 | grep OK
 
 tar -zxvf "${KREW_TGZ}"
 ./"krew-${KREW_ARCH}" install krew
-echo 'export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"' >> /home/tooling/.bashrc
+echo 'export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"' >> ${PROFILE_EXT}
 export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
 # kubens and kubectx
 kubectl krew install ns
@@ -419,16 +425,15 @@ RUN dnf -y install bash-completion \
     && rm -rf /var/cache/yum
 
 RUN <<EOF
-echo "source /etc/profile.d/bash_completion.sh" >> /home/tooling/.bashrc
 oc completion bash > /usr/share/bash-completion/completions/oc 
 kubectl completion bash > /usr/share/bash-completion/completions/kubectl
 cat ${NVM_DIR}/bash_completion > /usr/share/bash-completion/completions/nvm
 EOF
 
-# Add sdkman's init script launcher to the end of the .bashrc since we are not adding it on sdkman install
-# NOTE: all modifications to the .bashrc must happen BEFORE this step in order for sdkman to function correctly
-RUN echo 'export SDKMAN_DIR="/home/tooling/.sdkman"' >> /home/tooling/.bashrc
-RUN echo '[[ -s "$SDKMAN_DIR/bin/sdkman-init.sh" ]] && source "$SDKMAN_DIR/bin/sdkman-init.sh"' >> /home/tooling/.bashrc
+## Add sdkman's init script launcher to the end of ${PROFILE_EXT} since we are not adding it on sdkman install
+## NOTE: all modifications to ${PROFILE_EXT} must happen BEFORE this step in order for sdkman to function correctly
+RUN echo 'export SDKMAN_DIR="/home/tooling/.sdkman"' >> ${PROFILE_EXT}
+RUN echo '[[ -s "$SDKMAN_DIR/bin/sdkman-init.sh" ]] && source "$SDKMAN_DIR/bin/sdkman-init.sh"' >> ${PROFILE_EXT}
 
 
 # Create symbolic links from /home/tooling/ -> /home/user/
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -20,12 +20,7 @@ LABEL io.openshift.expose-services=""
 USER 10001
 
 # We install everything to /home/tooling/ as /home/user/ may get overriden, see github.com/eclipse/che/issues/22412
-RUN mkdir -p /home/tooling/
 ENV HOME=/home/tooling
-
-RUN cp /home/user/.bashrc /home/tooling/.bashrc
-# /home/user/.bashrc will be replaced with the a symlink from /home/tooling/.bashrc
-RUN rm /home/user/.bashrc    
 
 # Java
 RUN curl -fsSL "https://get.sdkman.io/?rcupdate=false" | bash \
@@ -439,10 +434,7 @@ RUN echo '[[ -s "$SDKMAN_DIR/bin/sdkman-init.sh" ]] && source "$SDKMAN_DIR/bin/s
 RUN chgrp -R 0 /home && chmod -R g=u /etc/passwd /etc/group /home
 
 # Create symbolic links from /home/tooling/ -> /home/user/
-RUN dnf -y install stow \
-    && dnf clean all \
-    && rm -rf /var/cache/yum \
-    &&  stow . -t /home/user/ -d /home/tooling/ --no-folding
+RUN stow . -t /home/user/ -d /home/tooling/ --no-folding
 
 
 # cleanup dnf cache
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -20,7 +20,7 @@ LABEL io.openshift.expose-services=""
 USER 10001
 
 # Java
-RUN curl -fsSL "https://get.sdkman.io" | bash \
+RUN curl -fsSL "https://get.sdkman.io/?rcupdate=false" | bash \
     && bash -c ". /home/user/.sdkman/bin/sdkman-init.sh \
              && sed -i "s/sdkman_auto_answer=false/sdkman_auto_answer=true/g" /home/user/.sdkman/etc/config \
 	     && sed -i "s/sdkman_auto_env=false/sdkman_auto_env=true/g" /home/user/.sdkman/etc/config \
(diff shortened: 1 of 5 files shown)


l0rd:
We should apply the same mechanism to the [base image](https://github.com/AObuchow/developer-images/blob/tooling_dir_new/base/ubi8/Dockerfile) too so that also the tools installed in that Dockerfile assume that `/home/tooling` is `$HOME`  (and there .bashrc should be created in /home/tooling).

It's surprising that `/home/user/.dotnet/tools /home/user/.krew/bin /home/user/.local/bin /home/user/.nvm/version /node/v16.14.0/bin /home/user/bin` are in PATH (I was expecting the `/home/tooling` counterpart)

> Something to note about using stow: the stow command in the entrypoint adds a bit of delay to the startup of the UDI, which is unfortunate. I haven't yet measured the exact amount of time it takes to startup on my local machine.

What if we run it in background so that it doesn't block the execution? The side effect may be that commands that follow will not find conf files in `/home/user` but we can fix the entrypoint to work fine. That makes me think that we can have problems with post run commands....

AObuchow:
> We should apply the same mechanism to the [base image](https://github.com/AObuchow/developer-images/blob/tooling_dir_new/base/ubi8/Dockerfile?rgh-link-date=2023-08-30T14%3A02%3A39Z) too so that also the tools installed in that Dockerfile assume that `/home/tooling` is `$HOME` (and there .bashrc should be created in /home/tooling).

We could either:
- set $HOME to `/home/tooling/` during the installation of tools in the base image
- set $HOME to `/home/tooling/` _and_ change the [useradd step](https://github.com/devfile/developer-images/blob/main/base/ubi8/Dockerfile#L94) to point to `/home/tooling/` instead of `/home/user/`.
    - The useradd command is creating the default `.bashrc` in `/home/tooling/` which we are copying over to /home/user/.
    - From what I understand, the useradd `-d` option is only supposed to create the home directory (if it doesn't exist) and set the user's login directory (which we override in our bashrc to point to the projects directory) 

Something to note about changing the base image to install to /home/tooling/: the `/home/user/.viminfo` and  `/home/user/.gitconfig` will be symbolic links located in `/home/tooling/. This means they won't persist if the user modifies them. This _should_ be okay, as users could just overwrite those files with automount configmaps in their workspaces. 


> It's surprising that `/home/user/.dotnet/tools /home/user/.krew/bin /home/user/.local/bin /home/user/.nvm/version /node/v16.14.0/bin /home/user/bin` are in PATH (I was expecting the `/home/tooling` counterpart)

`/home/user/.krew/bin /home/user/.local/bin /home/user/.nvm/version /node/v16.14.0/bin /home/user/bin` are on PATH because they're coming from the .bashrc and use the $HOME environment variable: when the .bashrc is sourced, $HOME is set to `/home/user/`.

```bash
# User specific environment
if ! [[ "$PATH" =~ "$HOME/.local/bin:$HOME/bin:" ]]
then
    PATH="$HOME/.local/bin:$HOME/bin:$PATH"
fi
export PATH
...
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
...
export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
```

I'm not exactly sure yet why `/home/user/.dotnet/tools` is on PATH as $HOME is set to `/home/tooling/` during the .NET installation step. It's also worth noting that `/home/user/.dotnet/` does not exist (even with this PR not applied). 


> > Something to note about using stow: the stow command in the entrypoint adds a bit of delay to the startup of the UDI, which is unfortunate. I haven't yet measured the exact amount of time it takes to startup on my local machine.
> 
> What if we run it in background so that it doesn't block the execution? The side effect may be that commands that follow will not find conf files in `/home/user` but we can fix the entrypoint to work fine. That makes me think that we can have problems with post run commands....

This is a good point - there is no guarantee that the entrypoint is run before postStart commands (see postStart documentation [here](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks)). This is very problematic IMO.

We could change the stow step to be a default devworkspace postStart command, but that also comes with its own complexity such as verifying we only add the stow postStart command when the UDI is used (and the UDI would not be fully usable outside of Che).

AObuchow:
My latest round of changes involve running the stow command in the actual Dockerfile itself (to prevent unnecessary extra startup time when `devEnvironments.persistUserHome` is not enabled) and modifying the UDI entrypoint to execute stow as a background process if /home/user/ has been mounted to by the PVC.

AObuchow:
> > It's surprising that `/home/user/.dotnet/tools /home/user/.krew/bin /home/user/.local/bin /home/user/.nvm/version /node/v16.14.0/bin /home/user/bin` are in PATH (I was expecting the `/home/tooling` counterpart)
> 
> `/home/user/.krew/bin /home/user/.local/bin /home/user/.nvm/version /node/v16.14.0/bin /home/user/bin` are on PATH because they're coming from the .bashrc and use the $HOME environment variable: when the .bashrc is sourced, $HOME is set to `/home/user/`.

@l0rd after some consideration, I believe it's best I modify any additions to the `.bashrc` so that `/home/tooling/` is hard-coded into PATH additions (rather than using $HOME). This way, if a postStart event invokes a command, they will be found on PATH regardless of whether the entrypoint has run or not. 


[`ea0cb13` (#115)](https://github.com/devfile/developer-images/pull/115/commits/ea0cb130f9a5bc7cf0a4dc1c8bbff003e72bd5a2) should probably be removed, as it causes more problems than it solves.

spuranam:
@AObuchow @l0rd  I tested the PR on DevSpaces v3.8, while it works as advertised, i also ran into oaccasions where not everything from /home/tooling would be end up /home/user running the following by manually after the devspace instance startup copies the missing items

```bash
cd ${DEVWORKSPACE_NAME}
stow . -t /home/user/ -d /home/tooling/ --no-folding --adopt
```

AObuchow:
@spuranam Thank you for the testing & feedback Satish. 

Would you be able to provide a specific example of a file(s) that isn't being linked from /home/tooling to /home/user? This would help me figure out edge-cases that I am currently missing. 

Also, are you trying to reference/use these missing files in a postStart command (and potentially running into a race-condition with the entrypoint)? 

Lastly, is there a specific reason why you're using the `--adopt` option when manually invoking `stow`?

spuranam:
@AObuchow The issue i am running into is that some times all the file /home/tooling would get linked sometime only some files would get linked to /home/user, for example i have seen many time /home/tooling/.sdkman and /home/tooling/.nvm will not get linked. 

My corrent workaround is to manually invoke `stow`, the only reason to `--adopt` is to force `stow` to run again since the run from the container entrypoint failed to link all the files, as without the command would fail.

I emailed all details needed for you to reproduce my results earlier today @l0rd , i am hoping he would be forward that email.

AObuchow:
I realized it's actually problematic to have the ` .viminfo` as a symbolic link, as it is [supposed to be an actual file by design](https://github.com/vim/vim/issues/2506#issuecomment-1154341891). I'll have to modify the stow command in the base container slightly to prevent this `.viminfo` from being stowed.

I'll have to update this PR as soon as I can with changes that address all the issues I've found so far.

universal/ubi8/Dockerfile
```diff
@@ -444,12 +441,35 @@ cd -
...
+# NOTE: all modifications to the .bashrc must happen BEFORE this step in order for sdkman to function correctly
+RUN echo 'export SDKMAN_DIR="/home/tooling/.sdkman"' >> /home/tooling/.bashrc
+RUN echo '[[ -s "$SDKMAN_DIR/bin/sdkman-init.sh" ]] && source "$SDKMAN_DIR/bin/sdkman-init.sh"' >> /home/tooling/.bashrc
+
 # Set permissions on /etc/passwd and /home to allow arbitrary users to write
 RUN mkdir -p /home/user && chgrp -R 0 /home && chmod -R g=u /etc/passwd /etc/group /home
 
+# Create symbolic links from /home/tooling/ -> /home/user/
+RUN stow . -t /home/user/ -d /home/tooling/ --no-folding
+
```
AObuchow, line 466: Need to remove this extra whitespace

AObuchow:
I have updated the PR with my latest changes and rebased it onto the main branch:
- The `.viminfo` is now being ignored by stow and being manually copied as an extra step. 
- I've ensured sdkman (and nvm) both load their init scripts from `/home/tooling/` to ensure they work properly. 
- The call to `stow` in the entrypoint is no longer being run as a background process to ensure it finishes successfully before the entrypoint exits.

There still are a few things that I need to address before this PR will no longer be in a draft state:
1. We should probably modify the `/etc/bashrc` instead of the `~/.bashrc` so that users are free to customize their bashrc without breaking important functionality.  This also lets us guarantee that `/etc/bashrc` will remain the same whenever the UDI is used (i.e. users won't be modifying or overwriting the /etc/bashrc). This is  important to ensure sdkman and nvm can be run as postStart commands.
2. I need to adopt the changes from https://github.com/devfile/developer-images/pull/117 to work with this PR so that the podman wrapper always works as expected.
3. I added a `chown -R 10001 ${HOME}/` in the [base image](https://github.com/devfile/developer-images/pull/115/files#diff-b6fe6f6f2726ba31be1fe707357ebb43670d415b1a001d0d0dbed214449ae09aR108) so that later calls to `chgrp -R 0 /home/tooling && chmod -R g=u /home/tooling` (like [here](https://github.com/devfile/developer-images/pull/115/files#diff-9f21bef4669d474b6ddef409447be53ae91c94cb50b6df9345149a0e89fb5391R40)) would work. Doing a recursive chown might be an antipattern, so we could instead chown the individual files.
4. Since `stow` is now being installed in the base image (and not the UDI), we probably will want to clean up the dnf cache as was [previously being done.](https://github.com/devfile/developer-images/pull/115/commits/29f7188f5420d3d827c76cc4224fe52ae6757b53#diff-9f21bef4669d474b6ddef409447be53ae91c94cb50b6df9345149a0e89fb5391L471-L472) 

I've pushed to `quay.io/aobuchow/universal-developer-image:stow` for testing, and [this devfile](https://github.com/AObuchow/udi-tooling/blob/master/devfile.yaml) can also be used.

AObuchow:
@spuranam Glad to hear my changes resolved those issues. I am currently working on fixing the issue with the podman wrapper.

@l0rd I've been trying to test my latest changes (pushed to `quay.io/aobuchow/universal-developer-image:stow-podman`) by replicating the testing instructions in https://github.com/devfile/developer-images/pull/107. However, even with the [factory link](https://workspaces.openshift.com/#https://github.com/l0rd/dockerfile-hello-world/tree/kubedock) from that PR, I am getting a service account issue on the DevSandbox & Dog Fooding cluster:
```
Error: pods is forbidden: User "system:serviceaccount:aobuchow-che-1cac83:workspacee2cbcbd7b10147e8-sa" cannot create resource "pods" in API group "" in the namespace "aobuchow-che-1cac83
```
I've made the following [devfile](https://github.com/AObuchow/udi-tooling-podman/blob/main/devfile.yaml#L14) to test out podman run with my latest changes. I also mounted a `/home/user/.kube/config` to my workspace (with garbage testing data) so that the podman.

l0rd:
@AObuchow I have tested your image and the wrapper is working as expected. 👍

AObuchow:
@l0rd I added two commits to address points **1.** and **3.** from my [earlier comment](https://github.com/devfile/developer-images/pull/115#issuecomment-1720169534). Point **4.** wasn't necessary since we're already [cleaning up the dnf cache](https://github.com/devfile/developer-images/blob/main/base/ubi8/Dockerfile#L26) after installing stow in the base image.

The newest changes have been pushed to `quay.io/aobuchow/universal-developer-image:stow`, which is used in [this devfile](https://github.com/AObuchow/udi-tooling/blob/master/devfile.yaml).

Something that came up while ensuring the `~/.bashrc` is no longer being modifed:

I noticed in the base image's [entrypoint](https://github.com/devfile/developer-images/blob/main/base/ubi8/entrypoint.sh#L8-L11), we're creating a `~/.bashrc` file manually (i.e. without `useradd`) in the event the `~/.bashrc` doesn't exist when the entrypoint is run. I'm under the impression that this is legacy code from https://github.com/devfile/developer-images/commit/5efac5706de63340b0ab768926367174386e3131 that was made obsolete when the [useradd step was added](https://github.com/devfile/developer-images/commit/8d9dc15a132768d2d95ecfe2f6895798c7621db7). 

However, [this is also present](https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/etc/entrypoint.sh#L45-L50) in the upstream UDI, so maybe this code is still necessary?

If it's not necessary, I could  remove it as part of this PR or in a future PR.

l0rd:
> However, [this is also present](https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/etc/entrypoint.sh?rgh-link-date=2023-09-20T05%3A29%3A09Z#L45-L50) in the upstream UDI, so maybe this code is still necessary?

You probably mean downstream right? In any case you are right. In the entrypoints (both upstream and downstream) we should assume `~/.bashrc` is there. So creation is not necessary ([I fixed a bug about that downstream](https://github.com/redhat-developer/devspaces-images/commit/06f2f5535ed73228c67f0b40d6905c3d30796cc9#diff-f507e1fba3158ecb25e3f6c3e8abe0bbf3b97f9c208d9239c6a00742e3334daaR49))

AObuchow:
> > However, [this is also present](https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/etc/entrypoint.sh?rgh-link-date=2023-09-20T14%3A08%3A08Z#L45-L50) in the upstream UDI, so maybe this code is still necessary?
> 
> You probably mean downstream right? In any case you are right. In the entrypoints (both upstream and downstream) we should assume `~/.bashrc` is there. So creation is not necessary ([I fixed a bug about that downstream](https://github.com/redhat-developer/devspaces-images/commit/06f2f5535ed73228c67f0b40d6905c3d30796cc9#diff-f507e1fba3158ecb25e3f6c3e8abe0bbf3b97f9c208d9239c6a00742e3334daaR49)).

Yes I meant the downstream UDI, my mistake. 

Sounds good, I added an extra commit to remove the `~/.bashrc` creation from the entrypoint.

AObuchow:
@l0rd Finally, this PR is ready for review. I've updated the PR description accordingly.

spuranam:
@AObuchow  @l0rd  I following two issues:

1. This [section in entrypoint timeout](https://github.com/spuranam/devspaces/blob/main/universal-developer-image/rootfs/entrypoint.sh#L40-L49), looking for `/home/user/.kube/config`, even though the `KUBECONFIG` environment variable is defned in the [Dockerfile](https://github.com/spuranam/devspaces/blob/main/universal-developer-image/Dockerfile#L105), thus devspace instance fails to starts.
2. If this variable [${KUBEDOCK_PARAMS}](https://github.com/spuranam/devspaces/blob/main/universal-developer-image/rootfs/entrypoint.sh#L55) is double quoted,  and when the coresponding env variable [KUBEDOCK_PARAMS](https://github.com/spuranam/devspaces/blob/main/parent-devfile.yaml#L35) in devfile, the entrypoint at runtime removes the white spaces between the parameters, to address i had to remove the double quotes around `${KUBEDOCK_PARAMS}` as mentioned ealier.

AObuchow:
@spuranam Thank you for testing & for the feedback.

I will apply the changes you made regarding point 2. to this PR.

For point 1. I haven't been able to reproduce this issue yet, even with `persistUserHome.Enabled: true` in the Che Cluster CR. However, I'm going to try and make a slight change that might help with the problem.

I will update this PR soon with the changes.

AObuchow:
@spuranam I applied your change for the KUBEDOCK_PARAMS and changed `$KUBECONFIG` to point to `/home/tooling/.kube/config`. The Che User Dashboard will prioritize creating the kube config at `$KUBECONFIG` (and if this is not set, it defaults to `$HOME/.kube/config`). 

Unfortunately, kubedock [does not respect the $KUBECONFIG environment variable](https://github.com/joyrex2001/kubedock/blob/d00924d693b7bb98e7d112f3b1d24d012055967e/cmd/server.go#L106-L112), and we need to supply the --kubeconfig argument to indicate where to find the kube config.

Since [you are defining](https://github.com/spuranam/devspaces/blob/main/parent-devfile.yaml#L35) `$KUBEDOCK_PARAMS`, you will have to also add `--kubeconfig $KUBECONFIG`.

I'm not sure if this change will actually resolve your issue, however, as I wasn't able to reproduce it yet. My current guess is that the dashboard is failing to [create the kube config](https://github.com/eclipse-che/che-dashboard/blob/35ffbe2fe1ebbc9ca37728a284264cd2c48b913e/packages/dashboard-backend/src/devworkspaceClient/services/kubeConfigApi.ts#L74-L84) in /home/user/.kube/ for some reason, though I am not sure why this would occur (mounting the PVC to /home/user/ is supposed to happen before the dashboard creates the kube config).

It might be worth checking the Che/DevSpaces-Dashboard logs to see if any errors are occuring, such as [this one](https://github.com/eclipse-che/che-dashboard/blob/35ffbe2fe1ebbc9ca37728a284264cd2c48b913e/packages/dashboard-backend/src/devworkspaceClient/services/kubeConfigApi.ts#L95).

I've pushed these latest changes to `quay.io/aobuchow/universal-developer-image:stow` and used [this devfile](https://github.com/AObuchow/udi-tooling-podman/blob/main/devfile.yaml) to test.

AObuchow:
I will have to make another change to the entrypoint so that it does not fail if the kube config is not found (instead, kubedock will just be disabled).

AObuchow:
@spuranam It might also be worth trying to extend the [length of the kubedock entrypoint timeout](https://github.com/spuranam/devspaces/blob/main/universal-developer-image/rootfs/entrypoint.sh#L42) to something like 30 seconds to see if that resolves the issue.

AObuchow:
I rebased this PR and added 2 extra commits:
- [`67b1cf4` (#115)](https://github.com/devfile/developer-images/pull/115/commits/67b1cf460ad5d1a7434ce7bbcf8a3a90be496c84) to ensure the container doesn't fail if `$KUBEDOCK_ENABLED=true` but the kube config could not be found
- [`b4df80f` (#115)](https://github.com/devfile/developer-images/pull/115/commits/b4df80f943799ebb64ab9588a6c0d43ffdeb2883) to allow extending the timeout when locating the kube config (though, this might not be necessary). The KUBEDOCK_TIMEOUT environment variable can be set in a devfile.

spuranam:
@AObuchow I can confirm that bumping timeout to 30 sec does indeed work, i also ported you latest change in to my custom UDI and now everthing works as expected.

AObuchow:
@spuranam awesome, I'm glad to hear everything is now working as intended. Thank you for the followup.

l0rd:
Leaving here the notes after our review of the PR:
- `/home/tooling/.kube/config` should not exist in the image
- we should execute `stow` once and skip it at workspace restart
- kubeconfig should be written in `/home/user` and never in `/home/tooling/`

AObuchow:
@l0rd I pushed 3 new commits to address the issues you mentioned:
- [`84d5d31` (#115)](https://github.com/devfile/developer-images/pull/115/commits/84d5d31be9ec5525f74375eb7bc6a92d2dc0eb62) ensures the kubeconfig will be created in /home/user/.kube/config
- [`0d963d9` (#115)](https://github.com/devfile/developer-images/pull/115/commits/0d963d903e63961c504a61cbeaad9a132997685d) ensures stow is only executed the first time the workspace is started up. Deleting `/home/user/.stow_completed` and running the entrypoint again allows re-running stow if ever needed.
- [`a0af6f2` (#115)](https://github.com/devfile/developer-images/pull/115/commits/a0af6f2e920cd41fd5c53106bc4f0e094baed991) was added to capture the output of stow, in case something goes wrong. 

[`0d963d9` (#115)](https://github.com/devfile/developer-images/pull/115/commits/0d963d903e63961c504a61cbeaad9a132997685d) could potentially be written more nicely, as I'm not super proficient in bash scripting. If you have any suggestions, feel free to propose them.

AObuchow:
There's another edge case I forgot to account for:
Since we are no longer moving the `.bashrc` from `/home/user/` to `/home/tooling/`, when persistUserHome is enabled, the `/home/user/.bashrc` will be overwritten and deleted.

I believe the solution is to move the `.bashrc` to `/home/tooling/`, add .bashrc to the stow-local-ignore and then manually copy it over so that changes to the` .bashrc` will persist between workspace startup and shutdown.

AObuchow:
I've pushed the (hopefully) last change for this PR & tested that everything seems to work fine with my [devfile](https://github.com/AObuchow/udi-tooling/blob/master/devfile.yaml) [1]. Changes were pushed to `quay.io/aobuchow/universal-developer-image:stow` 

[1] One weird thing: I noticed `/tmp/stow.log` was not actually created when the container started up even though the entrypoint ran successfully (and `~/.stow_complete` was created). Deleting `~/.stow_complete` and manually running the entrypoint resulted in `/tmp/stow.log` being created.

spuranam:
@AObuchow  @l0rd  i think i just found another case where this does not work, take this CR for example

```yaml
apiVersion: org.eclipse.che/v2
kind: CheCluster
metadata:
  name: devspaces
  namespace: openshift-devspaces
  annotations:
    argocd.argoproj.io/sync-wave: "3"
    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
spec:
  components:
    cheServer:
      debug: false
      logLevel: INFO
    dashboard: {}
    devWorkspace:
      runningLimit: "1"
    devfileRegistry:
      disableInternalRegistry: true
    imagePuller:
      enable: true
    metrics:
      enable: true
    pluginRegistry:
      openVSXURL: "https://open-vsx.org"
  containerRegistry: {}
  devEnvironments:
    secondsOfRunBeforeIdling: -1
    secondsOfInactivityBeforeIdling: 900
    disableContainerBuildCapabilities: false
    persistUserHome:
      enabled: true
    serviceAccount: devworkspace
    serviceAccountTokens:
      - name: bound-sa-token
        mountPath: /var/run/secrets/openshift/serviceaccount
        audience: openshift
        expirationSeconds: 3600
        path: token
    defaultComponents:
      - name: tools
        attributes:
          controller.devfile.io/merge-contribution: true
          pod-overrides:
            {
              "metadata":
                {
                  "annotations":
                    {
                      "io.kubernetes.cri-o.Devices": "/dev/fuse",
                      "io.openshift.podman-fuse": "",
                    },
                },
            }
        container:
          image: registry.xyz.com/devspaces/udi-ubi9@sha256:8f264db229d12f5bed253807df8b37a8af5aa6bea76f615c69fb8cbce849b46f
          env:
            - name: VSCODE_NODEJS_RUNTIME_DIR
              value: /usr/bin
            - name: TESTCONTAINERS_HOST_OVERRIDE
              value: tcp://127.0.0.1:2475
            - name: TESTCONTAINERS_RYUK_DISABLED
              value: "true"
            - name: TESTCONTAINERS_CHECKS_DISABLE
              value: "true"
            - name: SDKMAN_CANDIDATES_DIR
              value: /home/user/.sdkman/candidates
            - name: KUBEDOCK_TIMEOUT
              value: "30"
            - name: KUBEDOCK_ENABLED
              value: "true"
            - name: KUBEDOCK_PARAMS
              value: "--image-pull-secrets=image-pull-secret --initimage=registry.xyz.com/devspaces/kubedock@sha256:173eb6f14efb85a3d6bb8b1893609cb8e2faf3b2434e5541c71f1349c5ba2650 --request-cpu=100m,500m --request-memory=100Mi,500Mi --reverse-proxy"
          sourceMapping: /projects
          cpuLimit: 2000m
          cpuRequest: 500m
          memoryLimit: 2Gi
          memoryRequest: 512Mi
          mountSources: true
          endpoints:
            - name: kubedock
              targetPort: 2475
              protocol: https
              exposure: public
              secure: true
              attributes:
                urlRewriteSupported: true
    defaultNamespace:
      template: <username>-devspaces
      autoProvision: false
    storage:
      pvcStrategy: per-user
      perUserStrategyPvcConfig:
        claimSize: 5Gi
        storageClass: px-repl2-block
      perWorkspaceStrategyPvcConfig:
        claimSize: 5Gi
        storageClass: px-repl2-block
  gitServices:
    github:
      - endpoint: "https://github.xyz.com"
        secretName: github-oauth-config
        disableSubdomainIsolation: false
  networking:
    auth:
      gateway:
        configLabels:
          app: che
          component: che-gateway-config
```

Next start the default "empty workspace", workspace start and stow does run since file "/home/user/.stow_completed" yet the symbolic links are mssing ... 

adding the following to the section `spec.devEnvironments.defaultComponents` does not help.

```yaml
commands:
  ## https://github.com/spuranam/devspaces/pull/5
  # Required for running nvm & sdkman in postStart commands of child devfile
  - id: load-environment
    exec:
      component: tools
      commandLine: source /home/tooling/.bashrc
events:
  postStart:
    - load-environment
 ```

Howevere running the following once the devspace instance starts manually does address the issue

```bash
/home/tooling/.bashrc
stow . -t /home/user/ -d /home/tooling/ --no-folding --adopt
```

spuranam:
@AObuchow @l0rd I think i found an issue with `stow`, in that if were to start an "empty workspave" and this "cheCluster" CR 

```yaml
apiVersion: org.eclipse.che/v2
kind: CheCluster
metadata:
  name: devspaces
  namespace: openshift-devspaces
  annotations:
    argocd.argoproj.io/sync-wave: "3"
    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
spec:
  components:
    cheServer:
      debug: false
      logLevel: INFO
    dashboard: {}
    devWorkspace:
      runningLimit: "1"
    devfileRegistry:
      disableInternalRegistry: true
    imagePuller:
      enable: true
    metrics:
      enable: true
    pluginRegistry:
      openVSXURL: "https://open-vsx.org"
  containerRegistry: {}
  devEnvironments:
    secondsOfRunBeforeIdling: -1
    secondsOfInactivityBeforeIdling: 900
    disableContainerBuildCapabilities: false
    persistUserHome:
      enabled: true
    serviceAccount: devworkspace
    serviceAccountTokens:
      - name: bound-sa-token
        mountPath: /var/run/secrets/openshift/serviceaccount
        audience: openshift
        expirationSeconds: 3600
        path: token
    defaultComponents:
      - name: tools
        attributes:
          controller.devfile.io/merge-contribution: true
          pod-overrides:
            {
              "metadata":
                {
                  "annotations":
                    {
                      "io.kubernetes.cri-o.Devices": "/dev/fuse",
                      "io.openshift.podman-fuse": "",
                    },
                },
            }
        container:
          image: registry.xyz.com/devspaces/udi-ubi9@sha256:8f264db229d12f5bed253807df8b37a8af5aa6bea76f615c69fb8cbce849b46f
          env:
            - name: VSCODE_NODEJS_RUNTIME_DIR
              value: /usr/bin
            - name: TESTCONTAINERS_HOST_OVERRIDE
              value: tcp://127.0.0.1:2475
            - name: TESTCONTAINERS_RYUK_DISABLED
              value: "true"
            - name: TESTCONTAINERS_CHECKS_DISABLE
              value: "true"
            - name: SDKMAN_CANDIDATES_DIR
              value: /home/user/.sdkman/candidates
            - name: KUBEDOCK_TIMEOUT
              value: "30"
            - name: KUBEDOCK_ENABLED
              value: "true"
            - name: KUBEDOCK_PARAMS
              value: "--image-pull-secrets=image-pull-secret --initimage=registry.xyz.com/devspaces/kubedock@sha256:173eb6f14efb85a3d6bb8b1893609cb8e2faf3b2434e5541c71f1349c5ba2650 --request-cpu=100m,500m --request-memory=100Mi,500Mi --reverse-proxy"
          sourceMapping: /projects
          cpuLimit: 2000m
          cpuRequest: 500m
          memoryLimit: 2Gi
          memoryRequest: 512Mi
          mountSources: true
          endpoints:
            - name: kubedock
              targetPort: 2475
              protocol: https
              exposure: public
              secure: true
              attributes:
                urlRewriteSupported: true
    defaultNamespace:
      template: <username>-devspaces
      autoProvision: false
    storage:
      pvcStrategy: per-user
      perUserStrategyPvcConfig:
        claimSize: 5Gi
        storageClass: px-repl2-block
      perWorkspaceStrategyPvcConfig:
        claimSize: 5Gi
        storageClass: px-repl2-block
  gitServices:
    github:
      - endpoint: "https://github.xyz.com"
        secretName: github-oauth-config
        disableSubdomainIsolation: false
  networking:
    auth:
      gateway:
        configLabels:
          app: che
          component: che-gateway-config
```

I am not sure if there is way to incorporate these flines into CR

```yaml
commands:
  # Required for running nvm & sdkman in postStart commands of child devfile
  - id: load-environment
    exec:
      component: tools
      commandLine: source /home/tooling/.bashrc
events:
  postStart:
    - load-environment
```

However i can run these command manually after the "empty workspace" startups:

```bash
source /home/tooling/.bashrc
stow . -t /home/user/ -d /home/tooling/ --no-folding --adopt
```

Things seem to work as expected.

AObuchow:
@spuranam I will see if I can find a workaround for the empty workspace case - we might be able to set up a default devfile that includes the `load-environment` postStart command.

However, the `load-environment` postStart command should only be required to run certain commands, such as `nvm` & `sdk` in other postStart commands. Since you are starting an empty workspace (with no devfile) there won't be any postStart commands, unless I'm misunderstanding something?

Also, I'm not sure why manually running `stow . -t /home/user/ -d /home/tooling/ --no-folding --adopt` is required. The entrypoint of your tools container image should be running `stow` already.

spuranam:
@AObuchow My observations are since there is no "poststart"  i have seen random failure of where in stow is failing ... or it exists without linking all the files/folders ... i let me see if i can get other user reproduce this, i will upload /tmp/stog.log file

l0rd:
@spuranam @AObuchow If the problem is the missing support of `commands` and `events` in the CheCluster `..defaultComponents` we may create a dedicated issue as I think it's unrelated to this PR (and this git repo in general).

l0rd (approved):

l0rd:
I have done a few tests, in particular using kubedock and restarting a workspace and everything works for me.

AObuchow:
@l0rd Thank you for the testing and all your help with this PR :) 

If we merge https://github.com/devfile/developer-images/pull/132 before this PR, I will rebase this PR and make the [appropriate modifications](https://github.com/devfile/developer-images/pull/132#discussion_r1345859079) before squashing the fixup commits in this PR. Otherwise, I can just squash it so that it is ready for merging.

AObuchow:
> @AObuchow @l0rd I am not sure if this has anything to do with "command" and "event" two other user tested my custom UDI image they both see this error in /tmp/stow.log

Correct, this seems to be an issue with stow and the filesystem of your custom UDI image. 

> 
> ```shell
> Planning stow of package .... done
> Processing tasks...
> stow: ERROR: Could not create directory: .config (File exists)
> ```
> 

I'm not sure yet what's going on here yet. It seems that `/home/user/.config/` already exists, and this is causing stow to fail. However, in my testing, if `/home/user/.config/` exists as a directory, then running `stow . -t /home/user/ -d /home/tooling/ --no-folding -v 2 ` completes successfully, without giving `stow: ERROR: Could not create directory: .config (File exists)`.

I haven't looked into the `podman info` error yet.

AObuchow:
@l0rd I've squashed all the fixup commits from this PR for when we are ready to merge it

AObuchow:
I rebased my changes and made the appropriate modification for the [podman docker alias](https://github.com/devfile/developer-images/pull/115/commits/7ac79710c16454500116b51c37e3bdffc30e66de#diff-9f21bef4669d474b6ddef409447be53ae91c94cb50b6df9345149a0e89fb5391R199). This PR should be good to merge finally.

l0rd (approved):
LGTM

PR #117, opened 2023-09-04
chore: define user directory for binaries

vitaliy-guliy:
`/home/user/.local/bin` directory may contain binaries, that could be inaccessible when using VS Code tasks. 

For instance, this directory contains a link to `podman`. If I launch VS Code task, environment variables appear to be not properly/completely initialized and `/home/user/.local/bin` is not added to `PATH`.

One of the solutions is to create a directory for binaries and add it to the `PATH` variable in the Dockerfile directly.

Required for https://github.com/eclipse/che/issues/22292

### DIFF
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -79,6 +79,11 @@ ENV KUBECONFIG=/home/user/.kube/config
 
 USER 0
 
+# Define user direcotry for binaries
+RUN mkdir -p /home/user/.local/bin && \
+    chgrp -R 0 /home && chmod -R g=u /home
+ENV PATH="/home/user/.local/bin:$PATH"
+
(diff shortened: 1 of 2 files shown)


svor (approved):

l0rd:
cc @AObuchow because it has an impact on https://github.com/devfile/developer-images/pull/115

l0rd (approved):

AObuchow (approved):

universal/ubi8/Dockerfile
```diff
@@ -79,6 +79,11 @@ ENV KUBECONFIG=/home/user/.kube/config
 
 USER 0
 
+# Define user direcotry for binaries
```
AObuchow, line 82: Typo `direcotry` -> `directory`

PR #136, opened 2023-10-31
fix: Increase amount of opened files

vitaliy-guliy:
Creates `/etc/containers/containers.conf` configuration file to increase amount of opened files for `podman build / podman run`

Required for https://github.com/eclipse/che/issues/22292

### DIFF
--- a/universal/ubi8/Dockerfile
+++ b/universal/ubi8/Dockerfile
@@ -219,14 +219,8 @@ RUN mkdir -p /var/lib/shared/overlay-images /var/lib/shared/overlay-layers; \
 RUN mkdir -p "${HOME}"/.config/containers && \
    (echo '[storage]';echo 'driver = "vfs"') > "${HOME}"/.config/containers/storage.conf
 
-# Create `/etc/containers/containers.conf` configuration file
-# to increase amount of opened files
-#
-# [containers]
-# default_ulimits = [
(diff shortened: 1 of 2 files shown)


universal/ubi8/Dockerfile
```diff
@@ -219,6 +219,15 @@ RUN mkdir -p /var/lib/shared/overlay-images /var/lib/shared/overlay-layers; \
...
    (echo '[storage]';echo 'driver = "vfs"') > "${HOME}"/.config/containers/storage.conf
 
+# Create `/etc/containers/containers.conf` configuration file
+# to increase amount of opened files
+#
+# [containers]
+# default_ulimits = [
+#  "nofile=65535:65535",
+# ]
+RUN (echo '[containers]'; echo 'default_ulimits = ['; echo ' "nofile=65535:65535",'; echo ']') >> /etc/containers/containers.conf
```
AObuchow, line 229: This seems good to me, but could also be done with a COPY directive (though you might also have to include a chmod as is done [here](https://github.com/devfile/developer-images/commit/18a2377420c96d5893bc794da80a55deae6893ce#diff-b6fe6f6f2726ba31be1fe707357ebb43670d415b1a001d0d0dbed214449ae09aR95))

amisevsk (approved):

AObuchow (approved):

PR #137, opened 2023-11-01
add: cuda gpu capable images

codekow:
I want to add some work that has been done to add nvidia gpu capabilities into the developer images.

Original work here:
https://github.com/redhat-na-ssa/demo-ocp-gpu


codekow:
I admit this is a YOLO PR.

I'm not really clear on how you might want this PR structured as I have some ideas how I would refactor to make this a little more modular.

Looking for feedback around what liberties I can take without breaking your current workflows.

PR #143, opened 2024-01-12
add clo monitor exemption for release

Jdubrick:
All of the devfile repositories are currently being monitored on https://clomonitor.io/projects/cncf/devfile. As there are no releases planned for GitHub and the monitor is only tracking GitHub releases I am marking this as exempt on the monitor. The addition of the `.clomonitor.yaml` file allows this exemption to be properly tracked. Additionally it allows users who view the `developer-images` repo on the monitor to know that releases are going to `quay.io`.

fixes https://github.com/devfile/api/issues/1388

### DIFF
--- a/.clomonitor.yaml
+++ b/.clomonitor.yaml
@@ -1,0 +1,7 @@
+# CLOMonitor metadata file
+# This file must be located at the root of the repository
+
+# Checks exemptions
+exemptions:
+  - check: recent_release # Check identifier (see https://github.com/cncf/clomonitor/blob/main/docs/checks.md#exemptions)
+    reason: "There are no releases planned for GitHub. Releases for images are done on a rolling basis to quay.io."
--- a/.clomonitor.yaml
+++ b/.clomonitor.yaml
@@ -4,4 +4,4 @@
 # Checks exemptions
 exemptions:
   - check: recent_release # Check identifier (see https://github.com/cncf/clomonitor/blob/main/docs/checks.md#exemptions)
-    reason: "There are no releases planned for GitHub. Releases for images are done on a rolling basis to quay.io."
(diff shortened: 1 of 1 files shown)


.clomonitor.yaml
```diff
@@ -0,0 +1,7 @@
+# CLOMonitor metadata file
+# This file must be located at the root of the repository
+
+# Checks exemptions
+exemptions:
+  - check: recent_release # Check identifier (see https://github.com/cncf/clomonitor/blob/main/docs/checks.md#exemptions)
+    reason: "There are no releases planned for GitHub. Releases for images are done on a rolling basis to quay.io."
```
AObuchow, line 7: We could potentially be explicit here and state that `Releases for images are done on a rolling basis to quay.io at quay.io/devfile/base-developer-image:ubi8-latest and quay.io/devfile/universal-developer-image:ubi8-latest.`

Though this information is already present in the README and if we ever change the image tag we'd have to remember to update it here.

@svor @ibuziuk @l0rd WDYT?

AObuchow (approved):

AObuchow:
LGTM, thank you for the PR @Jdubrick. Will wait for other feedback and then this should be good to merge.

.clomonitor.yaml
AObuchow: @svor +1 that sounds good to me.

@Jdubrick When you have a moment, can you please change the `reason`field to @svor's suggestion please? Then we can get this merged. Thank you!

Jdubrick: @AObuchow all done!

AObuchow (approved):

AObuchow:
@Jdubrick thank you again for the PR!

PR #167, opened 2024-04-11
Dockerfiles: run tar x with --no-same-owner

nalind:
When we're extracting tar archives, always use --no-same-owner, so that ownership doesn't get set to oddball values that might have been used by the creator of those tar archives.

I'm running podman as an unprivileged user, and the default configuration can't build or pull images with content ownership UIDs or GIDs above 65536 or so.  Examples that I ran into in quay.io/devfile/universal-developer-image@sha256:2364d29270cdeadb15042785411900776a0b57dadc352befcff547242dfd8eb1 include /usr/local/bin/LICENSE (301071:301071) and /usr/local/bin/tkn (in two layers, one owned by 301071:64025, the other by 0:1000640000).

With this change applied, I can pull the built image without triggering an error.

### DIFF
--- a/base/ubi8/Dockerfile
+++ b/base/ubi8/Dockerfile
@@ -40,7 +40,7 @@ RUN \
     curl -sSLO "${GH_TGZ_URL}"; \
     curl -sSLO "${GH_CHEKSUMS_URL}"; \
     sha256sum --ignore-missing -c "gh_${GH_VERSION}_checksums.txt" 2>&1 | grep OK; \
-    tar -zxvf "${GH_TGZ}"; \
+    tar -zxv --no-same-owner -f "${GH_TGZ}"; \
     mv "gh_${GH_VERSION}_${GH_ARCH}"/bin/gh /usr/local/bin/; \
     mv "gh_${GH_VERSION}_${GH_ARCH}"/share/man/man1/* /usr/local/share/man/man1; \
     cd -; \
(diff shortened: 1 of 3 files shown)


AObuchow:
This looks sane to me upon quick inspection. Letting the CI tests run

AObuchow:
I also wonder if this is the root cause of https://github.com/eclipse-che/che/issues/22799

@nalind have you pushed the resulting image build from your changes anywhere? I'd like to quickly test something :)

nalind:
> I also wonder if this is the root cause of [eclipse-che/che#22799](https://github.com/eclipse-che/che/issues/22799)

Yes, this should prevent that error when pulling the image.

There's a subtle bug in how we handle parsing for heredocs in `podman build` that prevents it from building the image, and I hope to fix the root of that in the imagebuilder package today.

> @nalind have you pushed the resulting image build from your changes anywhere? I'd like to quickly test something :)

I've pushed it to quay.io/nalind/testing:udi.  Thanks!

AObuchow:
@nalind thank you SO much, I was really puzzled by this bug. Your fix looks good to me :)

AObuchow (approved):

nalind:
Thanks!  https://github.com/containers/buildah/pull/5473 aims to fix the other problem.

svor (approved):
lgtm

PR #171, opened 2024-05-06
Update dotnet  from 6.0 to 8.0

popalex:
Update dotnet version from 6.0 to 8.0, latest LTS version at this time.

Signed-off-by: Alex Popescu <pop.alx@gmail.com>

PR #172, opened 2024-05-10
Fix: pin stow to version 2.3.1

AObuchow:
Stow recently released version 2.4.0, which [does not allow](https://github.com/eclipse-che/che/issues/22957#issue-2285143021) stowing absolute symbolic links. As a temporary workaround, this PR downgrades stow back to version 2.3.1.

Stow is being downloaded from a GNU mirror, and then verified using the official GNU keyring (see the bottom of https://ftp.gnu.org/README). We remove the GNU gpg related files once we're done with them.

Unfortunately, the binary of stow is not available for direct usage, so stow is built and then moved to `/usr/bin/` so that it is available on $PATH to be used for the rest of the UBI & UDI builds (moving stow to `/usr/local/bin/` didn't seem to put it on $PATH).

Fixes https://github.com/eclipse-che/che/issues/22957

### DIFF
--- a/base/ubi8/Dockerfile
+++ b/base/ubi8/Dockerfile
@@ -25,7 +25,7 @@ RUN mkdir -p /home/tooling/
 RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && \
     dnf update -y && \
     dnf install -y bash curl diffutils git git-lfs iproute jq less lsof man nano procps p7zip p7zip-plugins \
-                   perl-Digest-SHA net-tools openssh-clients rsync socat sudo time vim wget zip stow && \
+                   perl-Digest-SHA net-tools openssh-clients rsync socat sudo time vim wget zip make && \
                    dnf clean all
 
 ## gh-cli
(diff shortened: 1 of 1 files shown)


svor (approved):
lgtm

AObuchow:
@ibuziuk @SDawley any thoughts on this? My only nitpick with this PR is the addition of `make` to the ubi. If there are no complaints, I'll have this merged as it's required to fix the UDI builds, and is a blocker for https://github.com/devfile/developer-images/pull/171

SDawley:
@AObuchow So long as this is the upstream solution, looks good to me.
curl won't work downstream though.

AObuchow:
> @AObuchow So long as this is the upstream solution, looks good to me. curl won't work downstream though.

Sounds good :) Downstream, we are manually [building stow from sources](https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/Dockerfile#L57-L66). Currently, it seems like the version of stow being [git cloned](https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/container.yaml#L71) downstream is [old](https://github.com/aspiers/stow/commit/4ef5eca4a9d107b24e712bb4c2c91f47e7e0fb85) so this issue hasn't arisen downstream. However, we might have to manually pin the commit to use for stow downstream if that commit is being automatically updated.

AObuchow:
Looks like I messed up when copying over the `/usr/share/` files :sweat_smile: as stow is still looking for them in `/tmp/.../build/share/`

image

Will look into this further tomorrow.

PR #179, opened 2024-07-24
Update ubi to configure kubedock and podman

dkwon17:
This PR adds podman/buildah and kubedock functionality to the ubi8 image.

~~To summarize the changes:~~
1. ~~In the dockerfile, download kubedock and create wrapper files around `dnf` and `podman`~~
2. ~~The `dnf` wrapper checks whether podman is installed. If it is installed, podman and kubedock are configured as it's done in the UDI8 image.~~

I had trouble building the images with podman, so the following instructions use docker, just like the contributing guide [here](https://github.com/devfile/developer-images/blob/5ede76af5cf41f617022c34d309d3daa2fc196fa/CONTRIBUTING.md)

To test:
1. Set env vars:
```
BASE_IMAGE=quay.io/<username>/<repo>:base-image
UDI=quay.io/<username/<repo>:udi
```

2. Build and push the base image
```
cd base/ubi8
DOCKER_BUILDKIT=1 docker image build --progress=plain -t $BASE_IMAGE .
docker push $BASE_IMAGE
```

3. Build and push the udi image
First, replace line 4 of the udi dockerfile [here](https://github.com/devfile/developer-images/blob/c909d75ed70fa268a497cd8f4caa192445c0c263/universal/ubi8/Dockerfile#L4) so that the base image matches the base image built from step 2.

Then, run the following:
```
cd universal/ubi8
DOCKER_BUILDKIT=1 docker image build --progress=plain -t $UDI .
docker push $UDI
```

4. Test by creating workspaces for the following cases

#### Case 1
The workspace starts with the base image `<CHE_HOST>/#https://github.com/dkwon17/empty?image=<BASE_IMAGE>`

#### Case 2
The workspace starts with the UDI image `<CHE_HOST>/#https://github.com/dkwon17/quarkus-api-example/tree/per-workspace?image=<UDI>`

Next, verify that the following command runs without any errors:
```
podman info
```

Run the `Package` and `Build image` devfile tasks to verify that podman image builds are also working.

Lastly, test kubedock by adding the `KUBEDOCK_ENABLED=true` env variable in the tooling container:
```
oc patch dw <workspace-name> --type='json' -p='[{"op": "add", "path": "/spec/template/components/0/container/env", "value": [{"name": "KUBEDOCK_ENABLED", "value": "true"}]}]' -n <namespace>
```

and by running the following when the workspace restarts:
```
$ podman run hello-world
!... Hello Podman World ...!

         .--"--.           
       / -     - \         
      / (O)   (O) \        
   ~~~| -=(,Y,)=- |         
    .---. /`  \   |~~      
 ~/  o  o \~~~~.----. ~~   
  | =(X)= |~  / (O (O) \   
   ~~~~~~~  ~| =(Y_)=-  |   
  ~~~~    ~~~|   U      |~~ 

Project:   https://github.com/containers/podman
Website:   https://podman.io
Desktop:   https://podman-desktop.io
Documents: https://docs.podman.io
YouTube:   https://youtube.com/@Podman
X/Twitter: @Podman_io
Mastodon:  @Podman_io@fosstodon.org
```


AObuchow:
@dkwon17 Looks good to me so far, will continue testing & reviewing on Monday :+1:

svor (commented):
tested by following steps from the description on dogfooding instance.
Everything works as expected except kubedock verification (second screenshot):
 
screenshot-che-dogfooding apps che-dev x6e0 p1 openshiftapps com-2024 07 29-18_14_31

screenshot-che-dogfooding apps che-dev x6e0 p1 openshiftapps com-2024 07 29-18_24_07

base/ubi8/dnf-wrapper.sh
```diff
@@ -0,0 +1,53 @@
...
+
+check_if_setup_needed() {
+    if command -v podman &> /dev/null && [ ! -e "$PODMAN_ORIGINAL_PATH" ]; then
+        return $TRUE
+    fi
+    return $FALSE
+}
+
+install_kubedock() {
+    curl -L https://github.com/joyrex2001/kubedock/releases/download/${KUBEDOCK_VERSION}/kubedock_${KUBEDOCK_VERSION}_linux_amd64.tar.gz | tar -C /usr/local/bin -xz --no-same-owner
```
AObuchow, line 25: In the case of a custom image that extends the UBI, I've run into the following issue:

```
Podman is installed, installing Kubedock...
/bin/dnf: line 25: KUBEDOCK_VERSION: unbound variable

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
Error: building at STEP "RUN sudo dnf install -y podman": while running runtime: exit status 2
```

Setting `ENV KUBEDOCK_VERSION 0.15.5` in the Dockerfile before doing `sudo dnf install -y podman` doesn't resolve the issue either, unfortunately.

I'm not entirely sure yet why the KUBEDOCK_VERSION environment variable can't be accessed from `dnf-wrapper.sh`? It seems like  $KUBEDOCK_VERSION is set correctly, as the following Dockerfile works:

```Dockerfile
FROM <universal-base-image-from-PR>

USER 0

RUN echo $KUBEDOCK_VERSION

RUN curl -L https://github.com/joyrex2001/kubedock/releases/download/${KUBEDOCK_VERSION}/kubedock_${KUBEDOCK_VERSION}_linux_amd64.tar.gz
```

But the following does not:
```Dockerfile
FROM <universal-base-image-from-PR>

USER 0

RUN  sudo dnf install -y podman

USER 1000
```

Something odd is going on with the dnf-wrapper.sh script when trying to expand `${KUBEDOCK_VERSION}`

AObuchow (commented):
Tested following the PR instructions & everything seems to work as expected :) Left some comments.

AObuchow:
> Everything works as expected except kubedock verification (second screenshot)

@svor did your workspace restart after patching in KUBEDOCK_ENABLED? i.e. `oc patch dw <workspace-name> --type='json' -p='[{"op": "add", "path": "/spec/template/components/0/container/env", "value": [{"name": "KUBEDOCK_ENABLED", "value": "true"}]}]' -n <namespace>`

In my testing, I got `export KUBEDOCK_ENABLED && ./entrypoint.sh` to work in enabling kubedock from the UDI.

svor:
Yes, I've restarted my workspace. I'll do another try tomorrow, maybe I've missed something 🤔

base/ubi8/dnf-wrapper.sh
dkwon17: It's working on the dogfooding cluster and my machine, could you share your base image?

AObuchow: My base image is `quay.io/aobuchow/universal-developer-image:david-kubedock-pr-base` and here's the problematic Dockerfile:

```
FROM quay.io/aobuchow/universal-developer-image:david-kubedock-pr-base

USER 0

RUN echo $KUBEDOCK_VERSION

RUN sudo dnf install -y podman

USER 1000
```

I built the base image as well as the above Dockerfile with podman. Maybe there's some podman bug going on here. I'll try re-building with Docker using your instructions.

AObuchow: I found my issue: I was invoking `dnf install -y podman` with `sudo` which was changing the bash environment. The following Dockerfile works as expected :grin: :

```Dockerfile
FROM quay.io/aobuchow/universal-developer-image:david-kubedock-pr-base

USER 0

RUN echo $KUBEDOCK_VERSION

RUN dnf install -y podman

USER 10001
```

svor (commented):
Works as expected
screenshot-nimbusweb me-2024 07 30-15_12_00

svor:
Maybe it makes sense to install podman into the base image
@dkwon17 @AObuchow @ibuziuk wdyt?

AObuchow:
I also prefer the simpler way to have podman in the base image, but an increase of the base image from ~150MB to ~280MB is roughly a 187% increase in size.

I don't know if we should assume that users (consumers of the UBI) will always want to be able to build containers. Kubedock is only ~30mb in size, so the majority of the increase in image size is caused by installing podman. IMO the current approach makes sense for a base image.

svor:
also it could be confused a bit that we have ENVs like PODMAN_WRAPPER_PATH and PODMAN_ORIGINAL_PATH in base image but no podman

AObuchow:
> also it could be confused a bit that we have ENVs like PODMAN_WRAPPER_PATH and PODMAN_ORIGINAL_PATH in base image but no podman

That's a good point.

I don't think we have an official page for universal developer (and base) image documentation in the Che Docs. When looking at the docs, I only see mentions of the UDI.

The change made by this PR might warrant adding a Che Docs page on the UBI & UDI, where we could mention that installing podman on the UBI will automatically install kubedock? There we could also mention the existence of the podman and dnf wrapper potentially, to clarify any confusion.

dkwon17:
Thank you for the feedback @AObuchow @svor 

I'm leaning towards the simple idea of adding podman into the base image to avoid adding the dnf wrapper logic which could be fragile, and to avoid possible side effects for images that build on top of it

AObuchow:
> I'm leaning towards the simple idea of adding podman into the base image to avoid adding the dnf wrapper logic which could be fragile, and to avoid possible side effects for images that build on top of it

I'm +1 for this. If consumers of the UBI complain about it's size, we always could go with the current approach of using a wrapper script for dnf.

svor (approved):
Changes look good
@dkwon17 am I right that each image that extends base image should have `source kubedock_setup` in entrypoint.json to enable kubedock? If it's true, maybe we could add this information into README?

dkwon17:
@svor Yes, that's correct, I will update the readme too for more details

dkwon17:
I've updated the README

svor (approved):

universal/ubi8/Dockerfile
```diff
@@ -43,8 +43,7 @@ RUN curl -fsSL "https://get.sdkman.io/?rcupdate=false" | bash \
              && sdk install maven \
              && sdk install jbang \
              && sdk flush archives \
-             && sdk flush temp" \
-    && chgrp -R 0 /home/tooling && chmod -R g=u /home/tooling
```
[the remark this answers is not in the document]
ibuziuk: looks like formatting

ibuziuk (approved):

AObuchow (approved):

PR #181, opened 2024-08-15
Remove unused /home/tooling/.local

dkwon17:
This PR fixes the issue where podman was not found when there is a volume mounted to /home/user/.local.

Fixes https://github.com/eclipse-che/che/issues/23095

In the base image, before this PR an empty `/home/tooling/.local` directory is created. Stowing this empty directory to `/home/user` created a hard link for the /home/.local directory, which caused both `/home/user/.local` and `/home/tooling/.local` to be affected by the volume mount to `/home/user/.local`.

This PR removes the `/home/tooling/.local` directory, since it was not used at all in the Dockerfile, and was only used in the entrypoint to store the podman wrapper. In this PR, the podman wrapper set up by the entrypoint is now stored in `/home/user/.local` instead.

### To test this PR

1. Build the base image:
```
$ cd base/ubi8
$ DOCKER_BUILDKIT=1 docker image build --progress=plain -t quay.io/<username>/<repo>:base-image .
```
I've built my own here: quay.io/dkwon17/developer-images:base-image

2. Push the image to quay.

3. Start the following workspace:
```
<CHE-HOST>#https://github.com/redhat-developer/devspaces/tree/devspaces-3-rhel-8?image=quay.io/<username>/<repo>:base-image
```
4. In the terminal, there should be no problems with running `podman info`
```
devspaces (devspaces-3-rhel-8) $ podman info
host:
  arch: amd64
  buildahVersion: 1.33.8
  cgroupControllers:
...
```

And `which podman` should output the following:
```
devspaces (devspaces-3-rhel-8) $ which podman
~/.local/bin/podman
```

### DIFF
--- a/base/ubi8/Dockerfile
+++ b/base/ubi8/Dockerfile
@@ -32,7 +32,8 @@ RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.n
 RUN dnf -y module enable container-tools:rhel8 && \
     dnf -y update && \
     dnf -y reinstall shadow-utils && \
-    dnf -y install podman buildah skopeo fuse-overlayfs
+    dnf -y install podman buildah skopeo fuse-overlayfs && \
+    dnf clean all
 
 ## gh-cli
 RUN \
@@ -97,12 +98,8 @@ RUN \
     cd - && \
     rm -rf "${TEMP_DIR}"
 
-
 # Define user directory for binaries
-RUN mkdir -p /home/tooling/.local/bin && \
-    chgrp -R 0 /home && chmod -R g=u /home && chown -R 10001 /home/tooling
 ENV PATH="/home/user/.local/bin:$PATH"
-ENV PATH="/home/tooling/.local/bin:$PATH"
 
 # Set up environment variables to note that this is
 # not starting with usernamespace and default to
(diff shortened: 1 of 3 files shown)


dkwon17:
To test this even further, you can build the UDI image with your custom base image by replacing line 4 of the udi dockerfile [here](https://github.com/devfile/developer-images/blob/c909d75ed70fa268a497cd8f4caa192445c0c263/universal/ubi8/Dockerfile#L4) so that the base image matches the base image built from this PR.

Then, the UDI can be built and pushed by running:
```
cd universal/ubi8
DOCKER_BUILDKIT=1 docker image build --progress=plain -t  quay.io/<username>/<repo>:udi .
docker push quay.io/<username>/<repo>:udi
```

To test that https://github.com/eclipse-che/che/issues/23095 is fixed, create a workspace with this URL:
```
<CHE-HOST>/#https://github.com/redhat-developer/devspaces/tree/devspaces-3-rhel-8?image=quay.io/<username>/<repo>:udi
```

I've built my own UDI image from this PR available here: `quay.io/dkwon17/developer-images:udi`

svor (approved):
Tested with https://github.com/redhat-developer/devspaces/tree/devspaces-3-rhel-8

AObuchow (commented):
LGTM, I was able to reproduce the bug with a workspace created from a fork of the `ds-plugin-registry-dev` devfile that uses a custom DWOC that disables persistent home on the dogfooding cluster.

With the fixed image, podman was again on path:
```
$ which podman
~/.local/bin/podman
```


However, with my custom DWOC, `podman info` failed:
```
$ podman info
ERRO[0000] running `/usr/bin/newuidmap 1520 0 1001030000 1 1 10000 65536`:  
Error: cannot set up namespace using "/usr/bin/newuidmap": fork/exec /usr/bin/newuidmap: operation not permitted
```

I'm almost certain the reason `podman info` is failing here is due to my custom DWOC having missing configuration fields, because:
- If I start the same workspace with persistent home enabled, the same issue arises (home persistence is not causing the issue)
- If I start the same workspace _without_ my custom DWOC, and use the dogfooding's Che-owned DWOC, then `podman info` succeeds.

Maybe this points to another bug? I'm not sure. I'm unable to see how the dogfooding's che-owned DWOC is configured (don't have the correct priviledges)

Here's my custom DWOC, it only has persisten home disabled:

```YAML
apiVersion: controller.devfile.io/v1alpha1
config:
  workspace:
    persistUserHome:
      enabled: true
kind: DevWorkspaceOperatorConfig
  name: custom-dwoc
  namespace: aobuchow-che-1cac83
```

dkwon17:
@AObuchow the custom DWOC should include:
```
config:
  workspace:
    containerSecurityContext:
      allowPrivilegeEscalation: true
      capabilities:
        add:
          - SETGID
          - SETUID
```

Podman build should be working after that

AObuchow:
@dkwon17 thanks for the follow-up, sounds good :)

PR #186, opened 2024-10-21
feat: Introduce multi-platform UDI image build

tolusha:
### Reference issue
https://github.com/eclipse-che/che/issues/23128


ibuziuk:
cc: @dkwon17 
need to coordinate this work with https://github.com/eclipse-che/che/issues/23034

PR #197, opened 2024-12-10
Fix home directory ownership so that podman 5.2.2 does not error

cgruver:
With the update of `podman` from 4.9.4 to 5.2.2 in UBI 9, podman fails to run if the ownership of $HOME/.config is incorrect.

This PR fixes that by moving the creation of $HOME and ${HOME}/.config/containers to the `ENTRYPOINT`.

### DIFF
--- a/base/ubi9/Dockerfile
+++ b/base/ubi9/Dockerfile
@@ -117,11 +117,6 @@ RUN mkdir -p /var/lib/shared/overlay-images /var/lib/shared/overlay-layers; \
     touch /var/lib/shared/overlay-images/images.lock; \
     touch /var/lib/shared/overlay-layers/layers.lock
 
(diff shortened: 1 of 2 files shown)


cgruver:
@ibuziuk   I created https://github.com/devfile/developer-images/pull/198 which also includes logic to detect support for fuse-overlayfs.

It also contains the changes from this PR, so they can either be processed in series, or this one can be closed if favor of the newer one.

ibuziuk:
@cgruver should we close this PR in favour of - https://github.com/devfile/developer-images/pull/198 ?

cgruver:
> @cgruver should we close this PR in favour of - #198 ?

Yes, #198 contains these changes as well.

PR #198, opened 2024-12-10
Detect support for fuse-overlayfs

cgruver:
Add logic to `entrypoint.sh` to detect support for `fuse-overlayfs`

### DIFF
--- a/base/ubi9/entrypoint.sh
+++ b/base/ubi9/entrypoint.sh
@@ -5,6 +5,13 @@ if [ ! -d "${HOME}" ]; then
   mkdir -p "${HOME}"
 fi
 
+# Configure container builds to use vfs
+if [ ! -d "${HOME}/.config/containers" ]
+then
+  mkdir -p ${HOME}/.config/containers
+  (echo '[storage]';echo 'driver = "vfs"') > "${HOME}"/.config/containers/storage.conf
+fi
+
 # Setup $PS1 for a consistent and reasonable prompt
 if [ -w "${HOME}" ] && [ ! -f "${HOME}"/.bashrc ]; then
   echo "PS1='[\u@\h \W]\$ '" > "${HOME}"/.bashrc
--- a/base/ubi9/entrypoint.sh
+++ b/base/ubi9/entrypoint.sh
@@ -5,11 +5,16 @@ if [ ! -d "${HOME}" ]; then
   mkdir -p "${HOME}"
 fi
 
-# Configure container builds to use vfs
+# Configure container builds to use vfs or fuse-overlayfs
 if [ ! -d "${HOME}/.config/containers" ]
 then
   mkdir -p ${HOME}/.config/containers
-  (echo '[storage]';echo 'driver = "vfs"') > "${HOME}"/.config/containers/storage.conf
+  if [ -c "/dev/fuse" ] && [ -f "/usr/bin/fuse-overlayfs" ]
+  then
+    (echo '[storage]';echo 'driver = "overlay"';echo '[storage.options.overlay]';echo 'mount_program = "/usr/bin/fuse-overlayfs"') > ${HOME}/.config/containers/storage.conf
+  else
+    (echo '[storage]';echo 'driver = "vfs"') > "${HOME}"/.config/containers/storage.conf
+  fi
 fi
 
 # Setup $PS1 for a consistent and reasonable prompt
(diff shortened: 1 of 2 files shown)


base/ubi9/entrypoint.sh
```diff
@@ -5,6 +5,18 @@ if [ ! -d "${HOME}" ]; then
   mkdir -p "${HOME}"
 fi
 
+# Configure container builds to use vfs or fuse-overlayfs
+if [ ! -d "${HOME}/.config/containers" ]
```
ibuziuk, line 9: @cgruver a matter of preference, but maybe you can change the formatting to follow the style applied in this file e.g. single line `if []; then`

cgruver: Sure.  Will do.

cgruver: I adjusted the formatting to match.

base/ubi9/entrypoint.sh
```diff
@@ -5,6 +5,16 @@ if [ ! -d "${HOME}" ]; then
   mkdir -p "${HOME}"
 fi
 
+# Configure container builds to use vfs or fuse-overlayfs
+if [ ! -d "${HOME}/.config/containers" ]; then
```
ibuziuk, line 9: @dkwon17 with this change we should probably update the docs as well, right?
e.g. https://docs.redhat.com/en/documentation/red_hat_openshift_dev_spaces/3.17/html-single/user_guide/index?extIdCarryOver=true&sc_cid=701f2000001Css5AAC#enabling-overlay-with-a-configmap
Now UDI will be smart enough to detect fuse automatically

ibuziuk:
verified on Developer Sandbox:

```
containers $ podman version
WARN[0000] Using cgroups-v1 which is deprecated in favor of cgroups-v2 with Podman v5 and will be removed in a future version. Set environment variable `PODMAN_IGNORE_CGROUPSV1_WARNING` to hide this warning. 
Client:       Podman Engine
Version:      5.2.2
API Version:  5.2.2
Go Version:   go1.22.7 (Red Hat 1.22.7-2.el9_5)
Built:        Wed Oct 23 09:08:59 2024
OS/Arch:      linux/amd64
containers $ cat storage.conf 
[storage]
driver = "vfs"
containers $ stat /dev/fuse
```

base/ubi9/entrypoint.sh
isuftin, line 9: Nitpicky, but couldn't the outer if block be removed and just be `mkdir -p "${HOME}/.config/containers"` ?

isuftin, line 9: What happens if persistent home is enabled and fuse-overlayfs gets disabled? Would `${HOME}/.config/containers/storage.conf` need to be rewritten? Would the outer `if` block stop that from happening?

cgruver: The only way to disable fuse-overlayfs would be to apply a machine-config to the cluster, so I would say it's not likely.  And if it did happen, that would be a disruptive change.

The reason that I put the first `if` block there is to not overwrite any existing podman config.  The user might have added addition config to storage.conf which we would not want to clobber on a workspace restart.

cgruver: The `if` block could be a bit more targeted to look specifically for storage.conf I suppose.

```bash
if [ ! -f "${HOME}/.config/containers/storage.conf" ]; then
```

But, there are valid configs where there is no `storage.conf` file, but other content under `${HOME}/.config/containers`.  Which is why I just check for the existence of the directory, not the file.

isuftin: @cgruver  that makes sense, thanks!

cgruver: __Note:__ As of OCP 4.15 `/dev/fuse` is available for leaking into a container OOTB.  No cluster change required.

With OCP 4.15 `/dev/net/tun` is also available.  

Pod annotation: `io.kubernetes.cri-o.Devices: "/dev/fuse,/dev/net/tun"`

So, to disable `fuse-overlayfs` support would require disruptive intervention on the part of cluster admins.

dkwon17: @ibuziuk , yes, a configmap will not be needed anymore if the tooling container image  is based on ubi/udi

dkwon17 (approved):

ibuziuk (approved):

PR #204, opened 2024-12-11
chore: bumping memoryLimit in devfile

ibuziuk:
can build the base, but not the universal on Developer Sandbox: 

Screenshot 2024-12-11 at 14 48 58


ibuziuk:
also, looks like there is a problem when building universal in the ephemeral mode on Developer Sandbox

``
Complete!
curl: (23) Failed writing received data to disk/application
subprocess exited with status 23
subprocess exited with status 23
``

PR #206, opened 2024-12-12
fix Entrypoint to use Base image entrypoint

cgruver:
This PR is a fix for https://github.com/eclipse-che/che/issues/23281

This PR also removes the run of `stow` during the build.  Running `stow` during the image build creates `/home/user/.config` resulting in `.config` being owned by root.

`.config` being owned by root breaks podman v5+


universal/ubi9/Dockerfile
```diff
@@ -413,18 +414,12 @@ EOF
 RUN echo 'export SDKMAN_DIR="/home/tooling/.sdkman"' >> ${PROFILE_EXT}
 RUN echo '[[ -s "$SDKMAN_DIR/bin/sdkman-init.sh" ]] && source "$SDKMAN_DIR/bin/sdkman-init.sh"' >> ${PROFILE_EXT}
 
-
-# Create symbolic links from /home/tooling/ -> /home/user/
-RUN stow . -t /home/user/ -d /home/tooling/ --no-folding
-
```
dkwon17, line 419: To be on the safer side, instead of removing the stow call, could we instead add `.config` in the stow ignore file [here](https://github.com/devfile/developer-images/blob/main/base/ubi9/.stow-local-ignore)?

We have the stow call here to move certain /home/tooling folders ([example](https://github.com/devfile/developer-images/blob/e2f4db7381e20438f6f30b7c8dd3dfbb62c366ce/universal/ubi9/Dockerfile#L77-L78)) to /home/user.

The stow call might not be entirely necessary since the binaries installed under /home/tooling are being added to the PATH anyways. I'll have to investigate more to be sure whether the stow call can be removed.

cgruver: Do we need the `stow` call during build?  I don't think that it does anything useful that isn't done when it runs in the `entrypoint.sh`

You'll see that exact same syntax is run in `entrypoint.sh`.  When it runs there, it runs as the user not root.

base/ubi9/entrypoint.sh
```diff
@@ -30,4 +30,25 @@ fi
...
+
+# /home/user/ will be mounted to by a PVC if persistUserHome is enabled
+mountpoint -q /home/user/; HOME_USER_MOUNTED=$?
+
+# This file will be created after stowing, to guard from executing stow everytime the container is started
+STOW_COMPLETE=/home/user/.stow_completed
+
+if [ $HOME_USER_MOUNTED -eq 0 ] && [ ! -f $STOW_COMPLETE ]; then
+    # Create symbolic links from /home/tooling/ -> /home/user/
+    stow . -t /home/user/ -d /home/tooling/ --no-folding -v 2 > /tmp/stow.log 2>&1
```
cgruver, line 44: @dkwon17 Does this call to `stow` not do the same thing that the call in the Dockefile does?

dkwon17: It does, but only if persistent home is enabled. We run stow in the image build to make links from /home/tooling to /home/user in case persistent home isn't enabled, so that files are still accessible under /home/user as expected

cgruver: Got it.  That makes sense.  I will revert that change and modify the stow ignore file.

universal/ubi9/Dockerfile
cgruver: @dkwon17 I restored the `stow` command in the Dockerfile.

dkwon17:
The build fails because the latest version of gopls is incompatible with the latest available version of go available in the `ubi-9-appstream-rpms` repo, I've made this PR to fix the gopls version to fix the build: https://github.com/devfile/developer-images/pull/207

cgruver:
> The build fails because the latest version of gopls (which was released yesterday) is incompatible with the latest available version of go available in the `ubi-9-appstream-rpms` repo, I've made this PR to fix the gopls version to fix the build: #207

I will close this PR and resubmit with the latest from `main`

cgruver:
Superseded by https://github.com/devfile/developer-images/pull/208

PR #221, opened 2025-05-22
chore: add workflow-dispatch event into ui9-build GH action

svor:
This PR adds support for manually triggering the UBI 9 base developer image build workflow via the GitHub UI

### DIFF
--- a/.github/workflows/ubi9-build.yaml
+++ b/.github/workflows/ubi9-build.yaml
@@ -4,6 +4,7 @@ name: Build of UBI 9 based Developer Images
 on:
   push:
     branches: [ main ]
+  workflow_dispatch:
 
   workflow_call:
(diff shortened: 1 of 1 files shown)


.github/workflows/ubi9-build.yaml
```diff
@@ -4,6 +4,7 @@ name: Build of UBI 9 based Developer Images
 on:
   push:
     branches: [ main ]
+  workflow_dispatch:
```
dkwon17, line 7: Maybe I just don't understand, but is it correct to have `workflow_dispatch:` with no value?

svor: I don’t think we need to define any inputs for triggering a new build manually, so the workflow_dispatch event can remain without any parameters.

PR #232, opened 2025-09-02
feat: Update /etc/sub*id files in order to run podman command in a us…

tolusha:
…er namespace


rohanKanojia (approved):

base/ubi9/entrypoint.sh
```diff
@@ -31,6 +31,49 @@ if ! whoami &> /dev/null; then
...
+  #         1       1001      64535
+  # For more details see:
+  # - https://man7.org/linux/man-pages/man7/user_namespaces.7.html
+  # - https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/
+  # - https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+  # - https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/
+  if [ -f /proc/self/uid_map ]; then
+    if ! grep -q '^\s*0\s\+0\s\+\([2-9]\|[1-9][0-9]\+\)' /proc/self/uid_map; then
+      echo "Running in a user namespace, user id: $(id -u)"
+      # By default, the valid UIDs/GIDs is the range 0-65535
```
dkwon17, line 53: IMHO the upstream https://github.com/devfile/developer-images should also work with CRI-O engine / OpenShift, can we support uids such as `1001250000`?

tolusha: We will use a dedicated SCC to enable container run capabilities [1]
So, the userid will always be lower than 65536

[1] https://github.com/cgruver/ocp-4-18-nested-container-tech-preview

base/ubi9/entrypoint.sh
```diff
@@ -31,6 +31,64 @@ if ! whoami &> /dev/null; then
...
+    # Even though, the range can be extended if needed by configuration, we can relay on the fact that IDs number at
+    # least 65536 are always available in a user namespace.
+    # See for more details:
+    # - https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-UserNamespaces
+
+    # To ensure the provided user ID stays within this range, `runAsUser` in the Pod spec should be set to a value below 65536.
+    # In OpenShift, the Container Security Context Constraint (SCC) should be created accordingly.
+    if [ "$(id -u)" -lt 65536 ]; then
+      START_ID=$(( $(id -u) + 1 ))
+      END_ID=$(( 65536 - ${START_ID} ))
```
dkwon17, line 79: Because of wording in https://man7.org/linux/man-pages/man5/subuid.5.html, could we rename to COUNT or some other name emphasizing that it is a count?

```suggestion
      COUNT=$(( 65536 - ${START_ID} ))
```

base/ubi9/entrypoint.sh
```diff
@@ -31,6 +31,64 @@ if ! whoami &> /dev/null; then
...
+    # For the mapping, podman refers to the /etc/sub*id files.
+    # See for more details:
+    # - https://man7.org/linux/man-pages/man5/subuid.5.html
+    # So, if user ID exceeds 65536 number, than it can be not be mapped, since it is not just available.
+    # As a result, podman commands fails.
+
+    # Even though, the range can be extended if needed by configuration, we can relay on the fact that IDs number at
+    # least 65536 are always available in a user namespace.
+    # See for more details:
+    # - https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-UserNamespaces
```
dkwon17, line 73:
```suggestion
    # Why do we need to update /etc/sub*id files?
    # On one hand, we are already in the user namespace, so there are at least 65536 UIDs/GIDs available.
    # For more details, see:
    # - https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/#id-count-for-each-of-pods
    # On the other hand, podman needs to create a new user namespace for any container being launched and map the container's user
    # and group IDs (UID/GID) to the corresponding user on the current namespace.
    # For the mapping, podman refers to the /etc/sub*id files.
    # For more details, see:
    # - https://man7.org/linux/man-pages/man5/subuid.5.html
    # So if the user ID exceeds 65535, it cannot be mapped if only UIDs/GIDs from 0-65535 are available.
    # If that's the case, podman commands would fail.

    # Even though the range can be extended using configuration, we can rely on the fact that there are at least 65536 user IDs available in a user namespace.
    # See for more details:
    # - https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-UserNamespaces
```

@tolusha could you please check if I understood this correctly?

PR #253, opened 2026-04-16
feat: adding gcloud CLI to ubi9 and ubi10

ibuziuk:
https://docs.cloud.google.com/sdk/gcloud 

````
ibuziuk@ibuziuk-mac google-cloud-sdk % cat LICENSE 
The Google Cloud CLI and its source code are licensed under Apache
License v. 2.0 (the "License"), unless otherwise specified by an alternate
license file.

You may obtain a copy of the License at

    http://www.apache.org/licenses/LICENSE-2.0

Note that if you use the Google Cloud CLI with any Google Cloud Platform
products, your use is additionally going to be governed by the license agreement
or terms of service, as applicable, of the underlying Google Cloud Platform
product with which you are using the Google Cloud CLI. For example, if you are
using the Google Cloud CLI with Google App Engine, your use would additionally
be governed by the Google App Engine Terms of Service.

This also means that if you were to create works that call Google APIs, you
would still need to agree to the terms of service (usually, Google's
Developer Terms of Service at https://developers.google.com/terms) for those
APIs separately, as this code does not grant you any special rights to use
the services.
````

### DIFF
--- a/universal/ubi10/Dockerfile
+++ b/universal/ubi10/Dockerfile
@@ -472,6 +472,41 @@ fi
 RUN curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/latest/skaffold-linux-${TARGETARCH} && \
     install skaffold /usr/local/bin/
 
+## Google Cloud CLI
+RUN <<EOF
+set -euf -o pipefail
+TEMP_DIR="$(mktemp -d)"
+cd "${TEMP_DIR}"
+
+GCLOUD_VERSION="512.0.0"
+
+case "$TARGETARCH" in
+    amd64)
+        GCLOUD_ARCH="x86_64"
+        ;;
+    arm64)
+        GCLOUD_ARCH="arm"
+        ;;
+    *)
+        echo "Skipping gcloud installation for unsupported architecture: $TARGETARCH"
+        exit 0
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by rohanKanojia — APPROVED** [review]

File: universal/ubi10/Dockerfile [hunk 1]
```diff
@@ -472,6 +472,41 @@ fi
 RUN curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/latest/skaffold-linux-${TARGETARCH} && \
     install skaffold /usr/local/bin/
 
+## Google Cloud CLI
+RUN <<EOF
+set -euf -o pipefail
+TEMP_DIR="$(mktemp -d)"
+cd "${TEMP_DIR}"
+
+GCLOUD_VERSION="512.0.0"
```
RomanNikitenko: minor: 
if I understand correctly - newer version is available
for example, in che-code we use version `563`: https://github.com/che-incubator/che-code/blob/main/build/dockerfiles/dev.Dockerfile#L23

ibuziuk: good, catch - it is actually `565` - https://docs.cloud.google.com/sdk/gcloud

